PR #103, opened 2018-05-15
added two configurable rolling log handlers

evaliula:


tabpy-server/tabpy_server/common/tabpy_logging.py
```diff
@@ -1,8 +1,37 @@
...
     def initialize(cls, logger):
         cls.logger = logger
+        cls.logger.setLevel = logging.getLevelName(LOG_LEVEL)
+
+        # create formatter
+        formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+        temp_dir = tempfile.gettempdir()
+        fh_info = handlers.RotatingFileHandler(filename=os.path.join(temp_dir, "tabpy_info.log",),
+                                               maxBytes=10000000, backupCount=5)
+        fh_info.setLevel(logging.INFO)
```
0golovatyi: Just to confirm - additionally to default logging with configurable log level with this change we are going to have log files with info and debug level in tmp folder?

0golovatyi: @evaliula Why do we need both - can we keep just the most verbose? Also, what is the purpose of user-defined logging level if we still to log with DEBUG/INFO?

evaliula: we can keep the most verbose one, I separate them because debug might have too much trace and it just a convenience if a parser is written or verification is based on INFO. If user only define logging info level , debug log file will not log debug trace.

evaliula:
updated to only one log handler

0golovatyi:
LGTM

tabpy-server/tabpy_server/common/config.py.template
```diff
@@ -2,4 +2,5 @@ TABPY_QUERY_OBJECT_PATH = '/tmp/query_objects'
 TABPY_PORT = 9004
 TABPY_SERVER_VERSION = 'Alpha'
 TABPY_STATE_PATH = './'
-TABPY_BIND_IP = '0.0.0.0'
\ No newline at end of file
+TABPY_BIND_IP = '0.0.0.0'
+TABPY_LOG_LEVEL = 'INFO'
```
johng42: Is the intent to change the level to INFO or should it be left at ERROR as the default?

Issue #108, opened 2018-06-08
Using Pipenv Instead of Conda / Pip


WillAyd:
In reading through the installation instructions I get the impression that this project is using anaconda simply to generate virtual environments, and not necessarily for its package management features as well, as it instructs users to still use pip for the actual installation.

As an alternate would you consider using Pipenv? I saw a talk on this at PyCon this year and it seems to have a bright future in the world of Python (it was made by the same guy that made the requests library):

https://docs.pipenv.org

The advantages of Pipenv here will be that:
 - Installation instructions will be simplified
 - You can distribute a Pipfile as part of the package to manage dependencies
 - It is arguably easier to use for activating / deactivating virtual environments

Happy to give it a go if you think its worthwhile

nmannheimer:
Hi @WillAyd,

If you're interested in exploring this we'd certainly be interested in seeing it set up!

Creating a containerized setup for TabPy is something we are considering, but it isn't on our immediate roadmap.

WillAyd:
Closed by #110

PR #110, opened 2018-06-28
Simplified Installation Instructions with Pipenv

WillAyd:
See #108 - this should greatly simplify the installation of packages by using Pipenv exclusively, instead of a mangled combination of conda / pip.

The user is of course still free to use whatever virtual environment and package configuration they'd like, but this should give simpler instructions to the 99% of users that just want something to work.


Pipfile
```diff
@@ -0,0 +1,21 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+future = "*"
+futures = "==3.1.1"
```
WillAyd: Typically wouldn't pin this requirement, but it is a bug in Pipenv:

https://github.com/pypa/pipenv/issues/1586

Not sure what the roadmap to support Py27 is so may not be a long term issue, but worth calling out

Pipfile
```diff
@@ -0,0 +1,21 @@
...
+tornado = "*"
+cloudpickle = "*"
+decorator = "*"
+python-dateutil = "*"
+genson = "*"
+jsonschema = "*"
+"809dc83" = {path = "./tabpy-client", edtiable = true}
+numpy = "*"
+
+[dev-packages]
```
WillAyd: One potential weakness here is that Pipenv doesn't support specifying multiple Python versions:

https://github.com/pypa/pipenv/issues/1050

If someone is trying to develop on a non-supported Python version this wouldn't stop them from doing so, but as far as distributions are concerned you'd be guarding against that in the `setup.py` requirements

Pipfile.lock
```diff
@@ -0,0 +1,196 @@
+{
```
WillAyd: In a pretty active development environment this file may be volatile, so I'll admit I'm not entirely sure of the consequences of including/excuding from VCS. 

That said, it is suggested by the Pipenv team to include so that's what I've done here:

https://github.com/pypa/pipenv/issues/598

server.md
```diff
@@ -2,121 +2,6 @@
...
-
-```
-### Installing TabPy
-
-Since TabPy is available on [PyPI](https://pypi.python.org/pypi/tabpy-server) you can install by simply running the following command.
-
-
-```Batchfile
-
-pip install tabpy-server
```
WillAyd: I thought this was pretty confusing before as it muddles the process for using the source code vs a pre-built distribution. As you'll see in the changes I've created a separate Contributing guide for dealing with the source and simply updated the `server.md` file to note how to deal with pre-buil distributions 

Side note - that itself is admittedly kind of strange in the current iteration - would probably be better served with a CLI component than having to launch directly from site-packages

Pipfile
sawdog: Have you considered baselining with pyenv with the pyenv-virtualenv plugin? This would accommodate a simplified install story as well as be a sole route to document running on all supported python versions? 

I've got a Dockerfile working that builds a running container that is likely useful to the project. I'd contribute that; I'd just need to tweak some minor modifications and likely switch to an alpine base image instead of the custom internal one I use now. I have this with Supervisor managing the processes. I was considering have it fronted with an Nginx proxy; that could be added to the example container or a single container of it's own to complete that story. For simplicity and 'access' to new users, the former would be more simple I'd imagine. Finally i'm fixing some "problems" with tabby.py startup that I found needed to be simplified that should be generally useful. I'll look further on Tableau's site or in the project for contributing and so on.

How actively is the main branch being maintained? 1 of your merges took a month and that looks only to mainly be pep8 cleanups......

WillAyd: I have not but at the same time I've never used that plug in. My reasoning for choosing pipenv was that it gives you both the virtual environment and package management all in one, and given the talk on it at PyCon this year I think its going to be well engrained in the future of the language. If you think this is better done with pyenv-virtualenv however you could always submit a PR to see what the team prefers :-)

I am not associated with Tableau in any way and therefore am not qualified to speak on the maintenance aspect, but I have gotten feedback on issues and PRs submitted so far

PR #115, opened 2018-07-10
LINTed psws folder

WillAyd:
Lot's of flakes here

**master**
```sh
$ flake8 | wc -l
     503
```

**more-flake8**
```sh
$ flake8 | wc -l
     379
```

### DIFF
--- a/tabpy-server/tabpy_server/psws/callbacks.py
+++ b/tabpy-server/tabpy_server/psws/callbacks.py
@@ -1,24 +1,24 @@
-import os
+import logging
 import sys
-import base64
 from time import sleep
-from tornado import gen
-from tornado.httpclient import AsyncHTTPClient
-from common.messages import (LoadObject, DeleteObjects, Msg,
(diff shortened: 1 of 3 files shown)


tabpy-server/tabpy_server/psws/callbacks.py
```diff
@@ -29,78 +29,80 @@ def wait_for_endpoint_loaded(py_handler, object_uri):
...
+        if object_type == 'alias':
+            msg = LoadObject(object_name, obj_info['target'],
+                             object_version, False, 'alias')
+        else:
+            local_path = object_path
+            msg = LoadObject(object_name, local_path, object_version,
+                             False, object_type)
+        py_handler.manage_request(msg)
 
-    except Exception as e:
```
WillAyd: AFAICT this try...except block had an uninitialized variable of `host_to_initialize` in the except clause, so if anything it would have just thrown another error and obfuscated what really happened

PR #132, opened 2018-09-22
Logging cleanup

### LINKED ISSUE #126: Remove Logging Module (?)
It's not entirely clear what the purpose of the logging module is in `tabpy-server`. I assume it's something that will eventually be integrated more tightly with Tableau's logging, but at the moment it's not quite there.

Would be preferable to create just one (or more) application wide logging mechanisms and import those when the server starts. The way the code base directly instantiates and manages imports now is needlessly complex.

https://docs.python.org/3/howto/logging-cookbook.html

WillAyd:
- [X] closes #126 

This should be a much more straightforward configuration for logging and has the added bonus of decoupling the client package from the server package (previously both would need a logging module placed in the server package). 

The __init__ files for the two packages are responsible for configuring the loggers with the appropriate namespacing. So long as any logger configured from there within the packages uses the convention of:

```python
```

They should inherit the configuration provided by the root of the package. Running the test suite confirms this yielding something as follows:

image


tabpy-server/tabpy_server/management/util.py
```diff
@@ -14,23 +12,25 @@
...
     Fill initial TabPy state from state file path
     '''
     try:
-        log_info("Loading state from state file")
+        logger.info("Loading state from state file")
         state_file_path = os.environ['TABPY_STATE_PATH']
+
+        # TODO: the below is namespaced with util which isn't valid (?)
+        # check if this is dead code and remove accordingly
         config = util._get_state_from_file(state_file_path)
```
WillAyd: It is stated in the comment above but unless I am overlooking something this code has to always error out since `util` is not a valid namespace in this context. This is probably dead code that can be removed altogether but that review and execution is suitable for a separate change

WillAyd:
@0golovatyi

0golovatyi:
@WillAyd  Actually looking at it your change can be merged without conflicts. I am going to look at it and get back to you in a day or two, feel free to ping me if there's no response in reasonable time.
And thank you for all the work - it is been a huge help for us.

tabpy-server/tabpy_server/__init__.py
```diff
@@ -1,4 +1,23 @@
-
-from ._version import get_versions
-__version__ = get_versions()['version']
-del get_versions
+import logging
+import logging.handlers
+import os
+import tempfile
+
+from ._version import get_versions
```
0golovatyi: Looks like you took some changes from our `dev` branch which are not merged to `master` yet. What we have in `master` is still uses versioneer code. I tried your changes (specifically the branch you are trying to merge from) and it doesn't work for me specifically for the versioning code.

WillAyd: Hmm OK. So is "dev" in reality the main branch?

0golovatyi: Well, it's complicated... :)

We meant to use dev when working on new features so we can have iterative changes without risking breaking master. And for other contributors, they should just create merge requests to merge in master. For when dev is out of sync with master we will take care of that.

The idea is you shouldn't deal with all the merging and conflict resolving for something which is not approved to be in master yet. Unless you want to :)

WillAyd: Hmm OK. Actually pulled this from master not dev so think that's a red herring, but I think it's a mistake on my end as to how I managed the merge conflicts with some of the larger changes being done recently. Should be able to look in more detail over next or two and resolve accordingly. Will ping back

WillAyd:
The caveat with reading from that file is going to be that we won't be able to un-bundle the client and server packages, which I *think* should be a goal as part of this. 

Maybe the server package reads from that file but the client package just uses default logging for now? Or reads an environment variable that is set outside of that configuration?

0golovatyi:
IMO both client and server packages using the same config file is something which needs to be changed. As they evolve they probably become completely independent things - client (rename to `tabpy_tools` in `dev`) will be a library and server is an app.

I believe it is better if logger in `tabpy_server` uses its own config file and `tabpy_client` uses config for the app it is used in.

For now hardcoded info level for client looks like good enough solution.

0golovatyi:
@WillAyd 
Merged.

PR #133, opened 2018-10-02
Fixes typo in readme

neilgoldman:
PyPy is an alternative implementation of Python. PyPi is the Python Package Index.


johng42:
Good catch.  Thanks!

PR #142, opened 2019-01-07
777044671

malakz:
Done


0golovatyi:
This is already covered as part of the agreement - https://help.github.com/articles/github-terms-of-service/.

PR #145, opened 2019-01-11
Bring the latest improvements

0golovatyi:
# This change includes the following improvements:

## Breaking changes

- The config file is now not just Python code but an actual config
- Tornado config file has different setting for CORS
- Setup scripts are deleted - setup (if needed) happens with startup script
- tabpy-client is tabpy-tools now

## Improvements

- Secure connection (HTTPS) is supported with Tableau 2019.2 and newer versions
- Documentation is improved with more examples added
- Versioning is done with Versioneer and github release tags
- Improved logging
- Unit tests are passing now 
- Configurations for Postman and Swagger are awailable to use those against running TabPy


tabpy-server/setup.py
```diff
@@ -25,21 +26,26 @@
...
-        'tornado',
-        'cloudpickle',
-        'decorator',
-        'python-dateutil',
-        'genson',
-        'jsonschema',
-        'tabpy_client',
-        'numpy'
-    ]
+        'tornado==5.1.1',
```
WillAyd: Does this really need to be pinned to this particular version? That's a rather strict requirement

tabpy-server/setup.py
```diff
@@ -25,21 +26,26 @@
...
-                  'tabpy_server': ['startup.*', 'state.ini.template']},
+                  'tabpy_server': ['startup.*', 'state.ini']},
     license='MIT',
     install_requires=[
+        'backports_abc',
+        'cloudpickle',
+        'configparser',
+        'decorator',
         'future',
         'futures',
```
WillAyd: If bumping to Py3 don't need this anymore, no?

tabpy-server/tabpy_server/_version.py
```diff
@@ -0,0 +1,520 @@
...
+# directories (produced by setup.py build) will contain a much shorter file
+# that just contains the computed version number.
+
+# This file is released into the public domain. Generated by
+# versioneer-0.18 (https://github.com/warner/python-versioneer)
+
+"""Git implementation of _version.py."""
+
+import errno
+import os
```
WillAyd: If requiring at least Python 3.4 should generally consider using `pathlib` instead of `os`

tabpy-server/tabpy_server/_version.py
```diff
@@ -0,0 +1,520 @@
...
+    # each be defined on a line of their own. _version.py will just call
+    # get_keywords().
+    git_refnames = "$Format:%d$"
+    git_full = "$Format:%H$"
+    git_date = "$Format:%ci$"
+    keywords = {"refnames": git_refnames, "full": git_full, "date": git_date}
+    return keywords
+
+
+class VersioneerConfig:
```
WillAyd: Empty class?

tabpy-tools/client_tests/test_client.py
```diff
@@ -6,14 +6,15 @@
 except ImportError:
     from mock import Mock
 
-from tabpy_client.client import *
+from tabpy_tools.client import *
```
WillAyd: Generally wildcard imports like this are frowned upon - any reason to not use explicit namespacing?

tabpy-tools/client_tests/test_rest.py
```diff
@@ -9,15 +8,14 @@
 
 import requests
 
+from tabpy_tools.rest import *
```
WillAyd: Similar namespacing comment

tabpy-tools/client_tests/test_rest.py
```diff
@@ -43,109 +41,109 @@ def mock_response(status_code):
...
     def test_GET(self):
         url = 'abc'
-        data = {'foo':'bar'}
+        data = {'foo': 'bar'}
         self.assertEqual(self.rnw.GET(url, data), 'json')
         self.rnw.session.get.assert_called_once_with(url,
-            params=data,
-            timeout=None)
+                                                     params=data,
+                                                     timeout=None)
```
WillAyd: Any reason for `timeout=None`? Assuming this uses requests behind the scenes that is the default anyway so no need to explicitly state, though it's probably best to actually provide some type of value to prevent these from hanging infinitely

tabpy-tools/tabpy_tools/client.py
```diff
@@ -20,10 +20,9 @@
 
 import os as _os
 
-
 import logging as _logging
```
WillAyd: Any reason for private namespacing here?

tabpy-tools/tabpy_tools/rest.py
```diff
@@ -1,21 +1,25 @@
 import logging as _logging
 from future.utils import with_metaclass
 import requests
+from re import compile as _compile
```
WillAyd: Same comment on private namespacing, though probably best just to import `re`

tabpy-tools/tabpy_tools/rest.py
```diff
@@ -1,21 +1,25 @@
...
+
 _logger = _logging.getLogger(__name__)
 
 from collections import MutableMapping as _MutableMapping
 
 import json as json
 
-from tabpy_server.common.tabpy_logging import (
+from .tabpy_logging import (
```
WillAyd: Generally I think you should get rid of the custom logging module as it doesn't serve a huge purpose. I have a separate PR for that

However, if you do feel a need to keep it around I think this import is incorrect and will be VERY buggy with some of the other path hackery going on in setup. Almost always preferable in a package of this size to use absolute imports

tabpy-tools/tabpy_tools/rest.py
```diff
@@ -322,18 +329,17 @@ def to_json(self):
 
     def __eq__(self, other):
         return (
-            type(self) == type(self)
-            and all((
-                getattr(self, a) == getattr(other, a)
-                for a in self.__rest__
-            )))
+                type(self) == type(self)
```
WillAyd: This looks like it should be `type(self) == type(other)` though I wonder if you even need to override this method in the first place?

tabpy-server/tabpy_server/common/endpoint_file_mgr.py
```diff
@@ -14,7 +14,7 @@
 import time
 from re import compile as _compile
 
-from tabpy_server.common.tabpy_logging import (
+from tabpy_tools.tabpy_logging import (
```
WillAyd: It's not explicitly stated in the setup but by importing like this you are essentially requiring that the client and server be installed together, which I don't think you really want to require (?)

startup.cmd
```diff
@@ -0,0 +1,73 @@
...
+python tabpy-server\utils\checkPrereqs.py
+IF %ERRORLEVEL% NEQ 0 (
+    GOTO:ERROR
+)
+
+
+REM Install requirements using Python setup tools.
+ECHO Installing any missing dependencies...
+CD %TABPY_ROOT%\tabpy-server
+python setup.py install>install.log 2>&1
```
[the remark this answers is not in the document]
0golovatyi: We actually thought about this and combining 2 scripts (setup and startup) in one has some benefits - users don't need to remember to run setup script every time something changes in Python environment. Running setup with every startup may be too excessive, but at least it guarantees setup step is always executed.

tabpy-server/server_tests/resources/valid.crt
```diff
@@ -0,0 +1,23 @@
+-----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
0golovatyi: These are included for added test cases only

tabpy-server/setup.py
0golovatyi: Testing this on different OSes and Python versions we found in some cases some other packages may bring an alpha version of Tornado (like 6a1) which causes failure at server starting due to breaking changes in Tornado 6+

CONTRIBUTING.md
```diff
@@ -1,33 +1,106 @@
...
+1. Open a windows command prompt.
+2. In the command prompt, navigate to the folder in which you would like to save
+   your local TabPy repository.
+3. In the command prompt, enter the following commands:
+
+    ```sh
+    git clone https://github.com/tableau/TabPy.git
+    cd TabPy
+    set TABPY_ROOT=%cd%
+    set PYTHONPATH=%PYTHONPATH%;%TABPY_ROOT%\tabpy-server;%TABPY_ROOT%\tabpy-server\tabpy_server;%TABPY_ROOT%\tabpy-tools;%TABPY_ROOT%\tabpy-tools\tabpy_tools;%TABPY_ROOT%\tabpy-server\tabpy_server\psws
```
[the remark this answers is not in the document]
0golovatyi: You are right, we don't need this. Fixing...

docs/tabpy-tools.md
```diff
@@ -173,8 +213,7 @@ client.query('clustering', x, y)
 
 Response:
 
-```
-
+```json
 {u'model': u'clustering',
```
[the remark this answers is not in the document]
0golovatyi: Fixing

tabpy-server/tabpy_server/_version.py
0golovatyi: This is the code versioneer generates and there's no point in editing it manually as it can be regenerated with any updates.

We are not locked to versioneer yet, it just looks like a best option for keeping version info in just one place (which is github release tag in case for versioneer).

tabpy-tools/client_tests/test_rest.py
0golovatyi: fixing

tabpy-tools/client_tests/test_rest.py
0golovatyi: Yeah... for the test code we will have more improvements sometime in future and for now, we just need to make sure existing tests whatever they are passing.

tabpy-tools/tabpy_tools/client.py
0golovatyi: You are right - we don't need this. I am removing all private namespaces in the file.

tabpy-tools/tabpy_tools/rest.py
0golovatyi: fixed

tabpy-tools/tabpy_tools/rest.py
0golovatyi: Good eye! Thank you for catching this!

We need to implement the methods for the type to be used as a dictionary key.

startup.cmd
```diff
@@ -0,0 +1,73 @@
+@ECHO off
+SETLOCAL
+
+
+REM Set environment variables.
+SET TABPY_ROOT=%CD%
+SET PYTHONPATH=%PYTHONPATH%;%TABPY_ROOT%;%TABPY_ROOT%\tabpy-server;%TABPY_ROOT%\tabpy-server\tabpy_server;%TABPY_ROOT%\tabpy-tools;%TABPY_ROOT%\tabpy-tools\tabpy_tools
```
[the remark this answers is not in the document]
0golovatyi: fixing

CONTRIBUTING.md
```diff
@@ -1,33 +1,106 @@
...
-pipenv install --dev
-```
+## Prerequisites
 
-This will create a dedicated virtual environment containing all of the required packages.
+These are prerequisites for an environment required for a contributor to
+be able to work on TabPy changes:
 
-## Starting the Server
+- Python 3.x
```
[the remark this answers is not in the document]
0golovatyi: We are currently looking at what versions range it makes sense and feasible to support (which includes testing TabPy for different OSes and Python versions). For now, I am going to leave the text as it is and in future when we narrow down the range the documentation and specs will be updated.

docs/server-startup.md
```diff
@@ -0,0 +1,121 @@
...
+- [Linux](#linux)
+  * [Command Line Arguments](#command-line-arguments-2)
+
+<!-- tocstop -->
+
+## Prerequisites
+
+To start up TabPy Server from an environment the following prerequisites are required:
+
+- Python 3 or greater
```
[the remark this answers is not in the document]
0golovatyi: See my responce above.

tabpy-server/setup.py
0golovatyi: You are right, removing `future` and `futures`

tabpy-server/tabpy_server/_version.py
0golovatyi: This is versioneer generated code, leaving it as it is.

startup.sh
```diff
@@ -0,0 +1,67 @@
...
+# Set environment variables
+echo Setting TABPY_ROOT to current working directory.
+export TABPY_ROOT=$PWD
+export PYTHONPATH="$PYTHONPATH:$TABPY_ROOT:$TABPY_ROOT/tabpy-server:$TABPY_ROOT/tabpy-server/tabpy_server:$TABPY_ROOT/tabpy-tools:$TABPY_ROOT/tabpy-tools/tabpy_tools"
+
+# Check for prerequisites
+#   - Python 3.X
+#   - Python modules: [sys, subprocess, setuptools, os ]
+echo Checking for prerequisites...
+python3 $TABPY_ROOT/tabpy-server/utils/checkPrereqs.py
```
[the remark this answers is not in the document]
0golovatyi: This script is an artifact for when we wanted to have separate setup/startup scripts and reusable requirement checking code :) Deleting it.

tabpy-tools/tabpy_tools/rest.py
0golovatyi: Again you are right and we are going to look at it and maybe remove the logger replacing it in the future. But for now, I am tending to keep it to not make this huge change list even bigger.

tabpy-server/tabpy_server/common/endpoint_file_mgr.py
0golovatyi: Client needs server and server needs client due to how the things are implemented right now :( But at least we remove tabpy-server package with this change.

WillAyd: Yea I think the logging was the only linkage between the two though. I should have time later this week to merge in this change and resolve conflicts so should be able to resolve the circular dependency thereafter

WillAyd:
@0golovatyi FYI I noticed that file permissions were changed on almost everything in the repo to be executable. Was this intentional?

I suppose there's nothing malicious to be executed standalone by any of the files, just don't typically see liberal permissions set like that. None of the files have shebangs anyway (?)

PR #156, opened 2019-01-28
Document supported configurations.

jnegara:
- Narrowed down Python support to 3.7.2
- Added living document for supported OS, TabPy server versions, and validators.


docs/server-startup.md
```diff
@@ -1,24 +1,24 @@
...
+  * [Command Line Arguments](#command-line-arguments-2)
+
+<!-- tocstop -->
+
 ## Prerequisites
 
 To start up TabPy Server from an environment the following prerequisites are required:
 
-- Python 3 or greater
+- Python 3.7.2
```
WillAyd: Just out of curiosity what was the decision to pin specifically to this version? I'm not aware of any other projects that make such a tight restriction down to a minor version, and most enterprise customers I would think may not be able to update to such a new version.

FWIW if deterministic builds is the goal here then I would suggest taking another look at Pipenv. It was recently removed with the major updates but would help with what I think you are trying to achieve here

jnegara: Hi Will! Good question. Currently we are listing Python 3.7.2 as a prerequisite because that is the version we are testing internally. We are currently in the process of determining a range of versions that we can commit to supporting.

We'll look into Pipenv, thanks for the suggestion!

WillAyd:
Sounds good. If looking for an outsider perspective I’d suggest requiring Python >= 3.6. That version introduced some pretty major language features that could help your code (specifically formatted string literals and typing) but has been out for two years, so strikes a balance with enterprises that may lag behind releases. 

If you wanted to go further back most packages I’m aware of only support 3.5 onwards, though there is a maintenance burden for sure of preceding 3.6 due to lack of aforementioned features

0golovatyi:
@WillAyd  We ended up with requiring Python 3.6.5. I spent some time and discovered for being able to use numpy 1.16 without significant (or rather any) code changes 3.6.5 is the lowest version which allows us to do that.

PR #157, opened 2019-01-29
Bug fix: install logs were only showing output from tabpy-tools install.

jnegara:
Updated startup scripts to include output from tabpy-server and tabpy-tools install.

### DIFF
--- a/startup.cmd
+++ b/startup.cmd
@@ -4,16 +4,21 @@ SETLOCAL
 
 REM Set environment variables.
 SET TABPY_ROOT=%CD%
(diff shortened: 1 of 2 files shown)


johng42:
Is there any way you can add a test for this change?

0golovatyi:
@johng42 It is not something we can cover with unit tests, specifically not with Python unit tests. If anything we should be able to have an integration test for starting TabPy when we have CI configured. So I am with @jnegara on this.

0golovatyi:
lgtm

PR #161, opened 2019-01-31
0.3.2 Release

0golovatyi:
- Remove versioneer and just replace it with VERSION file
- Require Python 3.6.5
- Require jsonschema to be compatible with 2.3.0
- Added setup instructions (known issues) for CentOS


docs/server-startup.md
```diff
@@ -96,6 +97,38 @@ The following is an example of how you might set both the port and the config:
...
+To install and enable all the required prerequisites follow the steps:
+
+```sh
+sudo yum update
+sudo yum groupinstall -y "Development tools"
+sudo yum install -y zlib-devel bzip2-devel openssl-devel ncurses-devel sqlite-devel readline-devel tk-devel gdbm-devel db4-devel libpcap-devel xz-devel expat-devel
+sudo yum install -y wget
+wget http://python.org/ftp/python/3.6.5/Python-3.6.5.tar.xz
+tar xf Python-3.6.5.tar.xz
+cd Python-3.6.5
```
WillAyd: Just out of curiosity is there any reason to suggest installing Python from source here? Seems simpler to let the package manager take care of that?

tabpy-server/server_tests/test_config.py
```diff
@@ -87,7 +87,7 @@ def test_config_file_present(self, mock_os, mock_path_exists, mock_psws, mock_sh
         mock_os.getenv.assert_has_calls(getenv_calls, any_order=True)
         self.assertEqual(settings['log_level'], 'WARNING')
         self.assertEqual(settings['port'], 1234)
-        self.assertEquals(settings['server_version'], tabpy_server.__version__)
+        self.assertEqual(settings['server_version'], open('VERSION').read().strip())
```
WillAyd: Any reason to move away from the dunder? PEP-8 actually encourages the former use and that's what I've come to expect with most pacakges

https://www.python.org/dev/peps/pep-0008/#id24

tabpy-server/setup.py
```diff
@@ -1,23 +1,27 @@
...
     from setuptools import setup
 except ImportError as err:
     print("Missing Python module requirement: setuptools.")
     raise err
 
+def read_version():
+    with open('../VERSION') as h:
+        return h.read().strip()
+
+__version__=read_version()
```
WillAyd: You might also want to have this in the package __init__ file as I guarantee external libraries may expect something like this to be available

tabpy-server/tabpy_server/version.py
```diff
@@ -0,0 +1,13 @@
+from pathlib import Path
```
WillAyd: Same comment - I think this would be better served in __init__

docs/server-startup.md
0golovatyi: Valid point. These steps are specifically for Python 3.6.5. I believe when installing with `yum install python36u` the latest (3.6.8) is installed. I'll update the documentation.

tabpy-server/tabpy_server/version.py
0golovatyi: I moved the code to `tabpy.py` instead

0golovatyi:
Updated the request

tabpy-server/setup.py
WillAyd: By library you mean Python package right? I interpret this repo as containing separate client and server packages, so I would think the most logical thing would be to have `__version__` defined in `__init__.py` for both of those packages.

I don't think it being in `tabpy.py` really does much as that's more of an executable module rather than an importable one, though certainly possible I'm misinterpreting intention here

tabpy-server/server_tests/test_config.py
```diff
@@ -21,6 +21,28 @@ def assert_raises_runtime_error(message, fn, args={}):
         assert err.args[0] == message
 
 
+def append_logger_settings_to_config_file(config_file):
```
WillAyd: Just as a tip you'd be much better off just using a StringIO or BytesIO instanced than an actual file for the tests. This could get rather wonky

tabpy-server/server_tests/test_config.py
```diff
@@ -70,24 +92,25 @@ class TestPartialConfigFile(unittest.TestCase):
     def test_config_file_present(self, mock_os, mock_path_exists, mock_psws, mock_shutil, mock_management_util,
                                  mock_tabpy_state, mock_parse_arguments):
         config_file = NamedTemporaryFile(delete=False)
-        config_file.write(b'TABPY_BIND_IP = 0.0.0.0\n'
-                          b'TABPY_QUERY_OBJECT_PATH = foo\n'
-                          b'TABPY_STATE_PATH = bar\n'
-                          b'TABPY_LOG_LEVEL = warning')
+        append_logger_settings_to_config_file(config_file)
+
+        config_file.write("[TabPy]\n"
```
WillAyd: So assuming you don't go through with my previous suggestion of using io instead of a file handle, you should at the very least probably use a context manager here for the tempfile

tabpy-server/server_tests/test_config.py
WillAyd: Does this actually pass? I would think the `open('VERSION')` expression should fail

tabpy-server/server_tests/test_config.py
```diff
@@ -127,128 +152,97 @@ def setUp(self):
         self.addCleanup(os.chdir, self.cwd)
 
     def test_http(self):
-        self.fp.write(b'TABPY_TRANSFER_PROTOCOL = http')
+        self.fp.write("[TabPy]\n"
+                      "TABPY_TRANSFER_PROTOCOL = http".encode())
         self.fp.close()
```
WillAyd: I guess unrelated to this PR but would be much better to call close in a teardown, or better yet use an io.StringIO instance which doesn't require explicit closing. If that doesn't work a still better option if you consider using other 3rd party testing tools like pytest could use a fixture to manage these tests.

tabpy-server/server_tests/test_config.py
```diff
@@ -127,128 +152,97 @@ def setUp(self):
...
 
-        settings, _ = get_config()
+        settings, _ = get_config(self.fp.name)
         self.assertEqual(settings['transfer_protocol'], 'http')
 
     def test_https_without_cert_and_key(self):
-        self.fp.write(b'TABPY_TRANSFER_PROTOCOL = https')
+        self.fp.write("[TabPy]\n"
```
WillAyd: Pretty strange to call `encode` on the string here before writing. Much better options include just creating bytes instead of strings or potentially not opening the file in binary mode (latter might be more complicated, didn't look comprehensively)

tabpy-server/setup.py
```diff
@@ -1,23 +1,21 @@
...
     print("Missing Python module requirement: setuptools.")
     raise err
 
-
 setup(
     name='tabpy-server',
-    python_requires='>=3.7.2',
-    version=versioneer.get_version(),
+    python_requires='>=3.6.5',
+    version=__version__,
```
WillAyd: This must also certainly fail right? I don't see where this is declared

tabpy-server/setup.py
```diff
@@ -36,7 +34,7 @@
         'decorator',
         'future',
         'genson',
-        'jsonschema>=2.3.0',
+        'jsonschema~=2.3.0',
```
WillAyd: What does unary inversion do here?

tabpy-server/tabpy_server/common/default.conf
```diff
@@ -1,11 +1,42 @@
...
 TABPY_BIND_IP = 127.0.0.1
-TABPY_LOG_LEVEL = INFO
 
 # To set up secure TabPy uncomment and modify the following lines.
 # Note only PEM-encoded x509 certificates are supported
 # TABPY_TRANSFER_PROTOCOL = https
 # TABPY_CERTIFICATE_FILE = path/to/certificate/file.crt
 # TABPY_KEY_FILE = path/to/key/file.key
+
+[loggers]
```
WillAyd: This is good but probably deserves a separate file (?) I don't think the startup script knows to read this

tabpy-server/tabpy_server/management/state.py
```diff
@@ -15,10 +15,8 @@
 
 from tabpy_server.management.util import write_state_config
 
-
```
WillAyd: Not sure if there is intention here but deleting these lines will typically cause LINTing tools to fail

tabpy-server/tabpy_server/tabpy.py
```diff
@@ -865,7 +849,7 @@ def set_parameter(settings_key, config_key, default_val=None, check_env_var=Fals
     # if state.ini does not exist try and create it - remove last dependence
     # on batch/shell script
     state_file_path = settings['state_file_path']
-    logger.info("Loading state from state file {}/state.ini".format(state_file_path))
+    logger.info("Loading state from state file {}".format(os.path.join(state_file_path, "state.ini")))
```
WillAyd: Secondary to this PR but should consider using pathlib operations instead of `os.path` going forward

tabpy-tools/setup.py
```diff
@@ -1,13 +1,31 @@
...
 try:
     from setuptools import setup
 except ImportError:
     from distutils.core import setup
 
+from pathlib import Path
+
+def read_version():
+    f = None
+    for path in ['VERSION', '../VERSION', '../../VERSION']:
```
WillAyd: Does offering three VERSION files really do anything? Seems rather fragile to do this

tabpy-server/tabpy_server/version.py
0golovatyi: also fixed in the way you suggested it should be

tabpy-server/server_tests/test_config.py
0golovatyi: I agree it looks weird but at the moment I am concentrated on making this work

tabpy-server/setup.py
0golovatyi: `~=` means `compatible with. We had issues with `>=` when one of the environments had a newer version of the package and Tornado started throwing exceptions complaining about missing attributes.

tabpy-server/tabpy_server/common/default.conf
0golovatyi: We actually want to move all the configurations (state.ini as well) to one config file for simplifying maintenance

tabpy-server/tabpy_server/tabpy.py
0golovatyi: Will fix in the future... :)

tabpy-tools/setup.py
0golovatyi: It is the same, but its location differs for when we run unit test, deploying package or running the app

PR #163, opened 2019-02-07
Update client.py

johng42:
make the default timeout to be 1000ms.  otherwise, we will complain that timeout cannot be zero


0golovatyi:
lgtm

PR #175, opened 2019-02-13
Dev pwd file

0golovatyi:
Work in progress...

Refactoring:
- Added application class to hold setting, state, credentials, etc.
- Cleaned up unit tests


tabpy-server/tabpy_server/handlers/management_handler.py
```diff
@@ -1,11 +1,47 @@
...
+from re import compile as _compile
+from uuid import uuid4 as random_uuid
+
+from tornado import gen
+
+from tabpy_server.handlers import MainHandler
+from tabpy_server.management.state import get_query_object_path
+from tabpy_server.psws.callbacks import on_state_change
+
+STAGING_THREAD = concurrent.futures.ThreadPoolExecutor(max_workers=3)
```
jnegara: I moved this var to [this file](https://github.com/tableau/TabPy/blob/dev/tabpy-server/tabpy_server/handlers/endpoint_handler.py#L13) but it likely doesn't belong there. Can you resolve this?

0golovatyi: fixed

PR #177, opened 2019-02-14
Bugfix: Startup scripts should fail when Python isn't in path.

jnegara:


WillAyd:
Were these not failing naturally when trying to execute python later in script?

jnegara:
> Were these not failing naturally when trying to execute python later in script?

@WillAyd They were. The updated scripts are more explicit about the cause of the failure.

PR #181, opened 2019-02-15
Add password file parameter and parsing (with unit tests)

0golovatyi:
Manual testing:
- starting up TabPy
- executing unit test


tabpy-server/server_tests/test_pwd_file.py
```diff
@@ -0,0 +1,107 @@
+import logging
+import os
+import unittest
+from argparse import Namespace
+from tempfile import NamedTemporaryFile
+
+from tabpy_server.app.app import TabPyApp
+from tabpy_server.app.ConfigParameters import ConfigParameters
+
+try:
```
WillAyd: Shouldn't need the try except anymore - unittest.mock has been in Python since 3.3

tabpy-server/server_tests/test_pwd_file.py
```diff
@@ -0,0 +1,107 @@
...
+from tabpy_server.app.ConfigParameters import ConfigParameters
+
+try:
+    from unittest.mock import patch, call
+except ImportError:
+    from mock import patch, call
+
+
+class TestPasswordFile(unittest.TestCase):
+    def __init__(self, *args, **kwargs):
```
WillAyd: Would be more idiomatic if you just put whatever you need in setUp

tabpy-server/server_tests/test_pwd_file.py
```diff
@@ -0,0 +1,107 @@
...
+        super(TestPasswordFile, self).__init__(*args, **kwargs)
```
WillAyd: While I would suggest getting rid of init altogether just calling out that you don't need to call the super in this fashion anymore as this is more of a Py2 compat call. Instead could just do `super().__init(*args, **kwargs)`

tabpy-server/server_tests/test_pwd_file.py
```diff
@@ -0,0 +1,107 @@
...
+        self.config_file = None
+        self.pwd_file = None
+        self.cwd = os.getcwd()
```
WillAyd: I would advise against using the current directory as it limits where the test suite can be executed from. Not a huge deal, but really only downsides to doing this

tabpy-server/server_tests/test_pwd_file.py
```diff
@@ -0,0 +1,107 @@
...
+        self.tabpy_cwd = os.path.join(self.cwd, 'tabpy-server', 'tabpy_server')
```
WillAyd: With minimum supported Python version I strongly advise using pathlib ops instead as they really simplify calls like this. This entire snippet could be emulated as `pathlib.Path.cwd() / 'tabby-server/tabpy_server'`

tabpy-server/server_tests/test_pwd_file.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def setUp(self):
+        os.chdir(self.tabpy_cwd)
+        self.config_file = NamedTemporaryFile(mode='w', delete=False)
```
WillAyd: What is the point of creating these files to subsequently close?

tabpy-server/server_tests/test_pwd_file.py
```diff
@@ -0,0 +1,107 @@
...
+        self.pwd_file.close()
+
+    def tearDown(self):
+        os.chdir(self.cwd)
+        os.remove(self.config_file.name)
+        self.config_file = None
+        os.remove(self.pwd_file.name)
+        self.pwd_file = None
+
+    def _set_file(self, file_name, value):
```
WillAyd: If you really wanted to do this probably easier to make a partial like `wopen = partial(open, mode='w')`

tabpy-server/server_tests/test_pwd_file.py
```diff
@@ -0,0 +1,107 @@
...
+        self._set_file(self.config_file.name,
+            "[TabPy]\n"
+            "TABPY_PWD_FILE = {}".format(self.pwd_file.name))
+        
+        self._set_file(self.pwd_file.name, "# just a comment")
+
+        with self.assertRaises(RuntimeError) as cm:
+            TabPyApp(self.config_file.name)
+            ex = cm.exception
+            self.assertEqual('Failed to read password file {}'.format(self.pwd_file.name), ex.args[0])
```
WillAyd: Cleaner to just use `self.assertRaises`

tabpy-server/server_tests/test_pwd_file.py
0golovatyi: good to know, will fix it

tabpy-server/server_tests/test_pwd_file.py
0golovatyi: ok

tabpy-server/server_tests/test_pwd_file.py
0golovatyi: fixing with the next push

tabpy-server/server_tests/test_pwd_file.py
0golovatyi: agreed

tabpy-server/server_tests/test_pwd_file.py
0golovatyi: fixing this as well

tabpy-server/server_tests/test_pwd_file.py
0golovatyi: The files need to be closed for their content to be available for a reader. The options were either I do it here and then in`_set_file` open the file for writing, or keep it open and close in `_set_file`

tabpy-server/server_tests/test_pwd_file.py
0golovatyi: Not sure what you mean. How the code will be different then?

tabpy-server/server_tests/test_pwd_file.py
0golovatyi: This is what happens here:
- L58: `self.assertRaises` checks `RuntimeError` is thrown on `TabPyApp` construction
- L61: check the exception has message we expect to see

0golovatyi:
@WillAyd Updated pull request

PR #185, opened 2019-02-16
Update /info response.

jnegara:
- Updated info JSON response. Now returns list of supported versions and features. i.e. with auth enabled, /info might return:
```json
{ 
   "description":"", 
   "creation_time":"0", 
   "state_path":"/Users/username/my-server-state-folder", 
   "server_version":"dev", 
   "name":"my-server-name", 
   "versions":{ 
      "v0.4":{ 
         "features":{ 
            "authentication":{ 
               "required":true, 
               "methods":{ 
                  "basic-auth":{} 
               } 
            } 
         } 
      } 
   } 
}
```
- Added unit tests for /info call in vanilla setting and with auth enabled.


jnegara:
@pep8speaks pep8ify

tabpy-server/server_tests/test_service_info_handler.py
```diff
@@ -0,0 +1,87 @@
+from tornado.testing import AsyncHTTPTestCase
+from tabpy_server.app import TabPyApp
+import simplejson as json
+from tabpy_server import __version__
+import tempfile, os
+
+
+def _create_expected_info_response(settings, tabpy_state):
+    response = {}
```
0golovatyi: you can combine declaration and initialization and even return, e.g.
```python
return {
    'description': tabpy_state.get_description(),
    'creation_time': tabpy_state.get_creation_time(),
    ...
}

File: tabpy-server/server_tests/test_service_info_handler.py [hunk 2]
```diff
@@ -0,0 +1,87 @@
...
+    response['name'] = tabpy_state.name
+    response['versions'] = settings['versions']
+    return response
+
+
+def _get_app_and_state(config_file=None):
+    tabpy_app = TabPyApp(config_file=config_file)
+    tornado_app = tabpy_app._create_tornado_web_app()
+
+    return tornado_app, tabpy_app.settings, tabpy_app.tabpy_state
```
0golovatyi: Caller get get settings and state with returned app, no need to return multiple parameters here

tabpy-server/server_tests/test_service_info_handler.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class TestServiceInfoHandlerDefault(AsyncHTTPTestCase):
+    def get_app(self):
+        app, self.settings, self.state = _get_app_and_state()
```
0golovatyi: Just keep `self.app`, no need in settings and state

tabpy-server/server_tests/test_service_info_handler.py
```diff
@@ -0,0 +1,87 @@
...
+        return app
+
+    def test_info(self):
+        """
+        Tests that /info returns the correct response on a vanilla TabPy instance.
```
0golovatyi: You can make test name something like `test_given_vanilla_tabpy_server_expect_correct_info_response` instead of adding docstring to a method :)

tabpy-server/server_tests/test_service_info_handler.py
```diff
@@ -0,0 +1,87 @@
...
+    def get_app(self):
+        test_pwd_file = tempfile.NamedTemporaryFile(prefix='___TestServiceInfoHandlerWithAuth_', suffix='.txt', delete=False)
+        test_pwd_file.write(b'username password')
+        test_pwd_file.close()
+
+        test_config = tempfile.NamedTemporaryFile(prefix='___TestServiceInfoHandlerWithAuth_', suffix='.conf',
+                                                  delete=False)
+        config = '[TabPy]\n' \
+                 'TABPY_PWD_FILE = {}\n' \
+                 'TABPY_STATE_PATH = ./tabpy-server/tabpy_server'.format(test_pwd_file.name)
```
0golovatyi: We don't want to use actual `state.ini` as it may contain user data or something which may interfere with tests. Instead, create temp state file and set an absolute path to it with the parameter. An example in `test_pwd_file.py` for how passwords files are created and used.

tabpy-server/server_tests/test_service_info_handler.py
```diff
@@ -0,0 +1,87 @@
...
+        test_config.write(bytes(config, 'utf-8'))
+        test_config.close()
+
+        app, self.settings, self.state = _get_app_and_state(test_config.name)
+
+        os.remove(test_config.name)
```
0golovatyi: We don't want to forget about those files in every single test so better create them in `setUp` and delete in `tearDown`

tabpy-server/tabpy_server/app/app.py
```diff
@@ -63,6 +62,25 @@ def __init__(self, config_file=None):
...
+        application = self._create_tornado_web_app()
+
+        init_model_evaluator(self.settings, self.tabpy_state, self.python_service)
+
+        if self.settings['transfer_protocol'] == 'http':
+            application.listen(self.settings['port'])
+        elif self.settings['transfer_protocol'] == 'https':
+            application.listen(self.settings['port'],
+                               ssl_options={
+                'certfile': self.settings['certificate_file'],
```
0golovatyi: SOmething wrong with alignment here

tabpy-server/tabpy_server/app/app.py
```diff
@@ -239,6 +242,9 @@ def set_parameter(settings_key,
                               self.settings[ConfigParameters.TABPY_PWD_FILE],
                               RuntimeError)
 
+        features = self._get_features()
+        self.settings['versions'] = {'v0.4': {'features': features}}
```
0golovatyi: Yeah... I don't like v0.4 here. Since it's not TabPy version but rather API version it should be less cryptic. IMO just `1` is the best option for now.

tabpy-server/server_tests/test_service_info_handler.py
johng42: We shouldn't expect unit tests to pass on a deployed instance of TabPy.  The tests should use the expected (default) state.ini - or have a separate integration test for that if we decide to not test it here.

tabpy-server/server_tests/test_service_info_handler.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        response = self.fetch('/info')
+        self.assertEqual(response.code, 200)
+        actual_response = json.loads(response.body)
+        expected_response = _create_expected_info_response(self.settings, self.state)
+
+        self.assertDictEqual(actual_response, expected_response)
+        self.assertTrue('versions' in actual_response)
+        versions = actual_response['versions']
+        self.assertTrue('v0.4' in versions)
```
johng42: Can we get the version from the VERSION file in the root folder?  

something like

from pathlib import Path

def get_tabpy_version():
    f = None
    for path in ['VERSION', '../VERSION', '../../VERSION']:
        if Path(path).exists():
            f = path

0golovatyi: @WillAyd we want it to be API version. IMO there's no need to have major.minor.patch for API version and we only increase version with breaking changes in API and for all optional feature like authorization, we have features for a version. And it should be possible in the future when we have multiple API versions to turn some of them off.

PR #187, opened 2019-02-19
better alert if Python can't be found.

johng42:


startup.cmd
```diff
@@ -6,7 +6,7 @@ REM Check for Python in PATH
 ECHO Checking for presence of Python in the system path variable.
 python --version >nul 2>&1
 IF %ERRORLEVEL% NEQ 0 (
-    ECHO     Cannot find Python.
+    ECHO     Cannot find Python.exe.  Check that Python is installed and is in the system PATH environment variable.
```
0golovatyi: We need similar change (after confirming it works) in `startup.sh`

PR #189, opened 2019-02-19
Update swagger config to parse new /info response schema.

jnegara:
Testing done:
- In swagger editor, manually verified the response schema is correct + generates the correct model.
- Verified it can parse the response from a local instance of TabPy server from the dev branch.

### DIFF
--- a/misc/TabPy.yml
+++ b/misc/TabPy.yml
@@ -182,6 +182,32 @@ definitions:
         type: string
       name:
         type: string
+      versions:
+        type: object
+        properties:
+          v1:
+            type: object
+            properties:
+              features:
+                type: object
+                properties:
+                  authentication:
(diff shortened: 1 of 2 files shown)


misc/TabPy.yml
```diff
@@ -1,241 +1,300 @@
...
-          type: object
-      
-externalDocs:
-  description: Find out more about TabPy
-  url: https://github.com/tableau/TabPy
\ No newline at end of file
+swagger: '2.0'
+info:
+  description: TabPy API Swagger
+  version: v0.4
```
0golovatyi: This should be v1 as it is for API version

0golovatyi:
Why the whole file shows as changed? Line ending? Need to know what causes this behavior.

PR #192, opened 2019-02-20
Dev validate creds

0golovatyi:
- Code to check if validation for creds needs to be done
- Certs validation implementation
- Unit tests


CONTRIBUTING.md
```diff
@@ -88,3 +88,22 @@ Access-Control-Allow-Methods = GET, OPTIONS, POST
 - Open `misc/TabPy.yml` in Swagger editor.
 - In case your TabPy server runs not on `localhost:9004` update
   `host` value in `TabPy.yml` accordingly.
+
+## Code styling
+
+On github repo for merge request `pycodestyle` is used to check Python code against our
+style conventions. You can run install it and run locally for file where modifications
```
johng42: typo 
"You can install it and run  ..."

tabpy-server/tabpy_server/app/app.py
```diff
@@ -320,7 +320,7 @@ def _parse_pwd_file(self):
                             'in password file'.format(row))
                         return False
 
-                    login = row[0]
+                    login = row[0].lower()
```
johng42: Do we need to document this (that names are converted to lower and not unique based on capitalization)?
And that we will use an English locale to perform the lower() operation...

CONTRIBUTING.md
0golovatyi: I actually changing the steps in upcoming PR

PR #193, opened 2019-02-21
Fix generic error message from startup.sh; add cli arg for using startup.sh in a test env.

jnegara:
These changes probably shouldn't be a part of the same PR.

- Added a parameter to check_status in startup.sh that allows you to add a specific error message.
- Made variables local in startup.sh.
- Added a CLI arg to startup.sh for skipping server startup for use in CI.
- Changed CLI arg format for config && updated documentation.
- Removed references to port CLI arg from documentation (as we removed it in a previous PR).


startup.sh
```diff
@@ -2,7 +2,7 @@
 
 function check_status() {
     if [ $? -ne 0 ]; then
-        echo TabPy startup failed.  Cannot find Python.  Check that Python is installed and is in the system PATH environment variable.
+        echo TabPy startup failed.
```
0golovatyi: How about providing message parameter for the function?

docs/server-startup.md
```diff
@@ -68,29 +68,20 @@ For example, in this case your config file would be located at ```%TABPY_ROOT%\t
 
 ### Command Line Arguments
 
-- To specify the *port* on which your server instance listens, set the ```-p``` command line argument as follows:
+- To specify the *config file* with which to configure your server instance, set the ```--config=*``` or ```-c=*``` command line argument as follows:
```
sbabayan: Did we remove the ability to set port independently of changing it in the config and setting your config on startup?

johng42: yes -we only parse for the config file.  see tabpy-server\tabpy_server\app\app.py

jnegara: We removed the port CLI arg with [this change](https://github.com/tableau/TabPy/commit/0cc57b9901be5829a1e574a98da37925387cf9b8#diff-d0486b0d7b181591f52bad014cd3a06f).

PR #194, opened 2019-02-22
Dev validate creds

0golovatyi:
Add basic unit tests to HTTP handlers for auth scenario


tabpy-server/server_tests/test_endpoint_handler.py
```diff
@@ -0,0 +1,111 @@
...
+        cls.pwd_file = tempfile.NamedTemporaryFile(
+            mode='w+t', prefix=prefix, suffix='.txt', delete=False)
+        cls.pwd_file.write('username {}'.format(
+            hashlib.sha3_256('password'.encode('utf-8')).hexdigest()))
+        cls.pwd_file.close()
+
+        # create state.ini dir and file
+        cls.state_dir = tempfile.mkdtemp(prefix=prefix)
+        cls.state_file = open(os.path.join(cls.state_dir, 'state.ini'), 'w+')
+        cls.state_file.write('[Service Info]\n'
```
jnegara: We do this in multiple tests. Maybe we should extract it into utils, especially as we continue to add tests for endpoint handlers.

tabpy-server/tabpy_server/handlers/base_handler.py
```diff
@@ -47,15 +51,44 @@ def _add_CORS_header(self):
...
+            self.request.headers,
+            "v1",
+            self.settings,
+            self.credentials)
+
+    def fail_with_not_authorized(self):
+        '''
+        Prepares server 401 response.
+        '''
+        logger.error('Failing with 401 for anothorized request')
```
jnegara: unauthorized

tabpy-server/server_tests/test_endpoint_handler.py
0golovatyi: agree, do it with your next change

tabpy-server/tabpy_server/handlers/base_handler.py
0golovatyi: fixed

PR #202, opened 2019-03-11
Cleanup Unneeded Modules in Setup

WillAyd:
Lot's of cruft in here

### DIFF
--- a/tabpy-server/server_tests/test_config.py
+++ b/tabpy-server/server_tests/test_config.py
@@ -3,10 +3,7 @@ import unittest
 import logging
 import tabpy_server
 
-try:
-    from unittest.mock import patch, call
-except ImportError:
-    from mock import patch, call
+from unittest.mock import patch, call
 
 from tempfile import NamedTemporaryFile
 from tabpy_server.tabpy import get_config, validate_cert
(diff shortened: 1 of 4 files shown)


tabpy-tools/tabpy_tools/rest.py
```diff
@@ -239,7 +238,7 @@ def __init__(self, name, bases, dict):
                 self.__rest__.add(k)
 
 
-class RESTObject(with_metaclass(_RESTMetaclass, _MutableMapping)):
+class RESTObject(_MutableMapping, metaclass=_RESTMetaclass):
```
WillAyd: The intention here is very difficult to reason about, but that's an artifact of existing code so doing minimum here to remove third party dependency

tabpy-server/server_tests/test_config.py
```diff
@@ -3,10 +3,7 @@
 import logging
 import tabpy_server
 
-try:
-    from unittest.mock import patch, call
-except ImportError:
-    from mock import patch, call
+from unittest.mock import patch, call
```
WillAyd: Given bump to Py3.6 don't need to try...except here. There's lots of other compat code that can be removed in subsequent PRs

0golovatyi:
@WillAyd we actually had to restore most of the deleted dependencies - even they are supposed to go with Python we immediately discovered Linux flavors where it is not the case :(

WillAyd:
Not sure I follow - do you have an example of that? 

Most of these are Py2 packages so they'd essentially provide noops

0golovatyi:
@johng42 Could you provide examples for Will's question?

PR #205, opened 2019-03-12
Flake8 of tabpy-tools

WillAyd:
To improve styling and reduce bugs. Ideally will have something like autopep8 set up in the future to catch these as part of CI

### DIFF
--- a/tabpy-tools/tabpy_tools/client.py
+++ b/tabpy-tools/tabpy_tools/client.py
@@ -21,7 +21,7 @@ import logging
 
 logger = logging.getLogger(__name__)
 
-_name_checker = compile(r'^[a-zA-Z0-9-_\ ]+$')
+_name_checker = compile(r'^[\w -]+$')
 
 if sys.version_info.major == 3:
     unicode = str
--- a/tabpy-tools/tabpy_tools/client.py
+++ b/tabpy-tools/tabpy_tools/client.py
@@ -21,7 +21,7 @@ import logging
 
 logger = logging.getLogger(__name__)
 
-_name_checker = compile('^[a-zA-Z0-9-_\ ]+$')
+_name_checker = compile(r'^[a-zA-Z0-9-_\ ]+$')
 
 if sys.version_info.major == 3:
     unicode = str
@@ -37,11 +37,13 @@ def _check_endpoint_type(name):
 
 def _check_hostname(name):
     _check_endpoint_type(name)
-    hostname_checker = compile('^http(s)?://[a-zA-Z0-9-_\.]+(/)?(:[0-9]+)?(/)?$')
+    hostname_checker = compile(
+        r'^http(s)?://[a-zA-Z0-9-_\.]+(/)?(:[0-9]+)?(/)?$')
 
     if not hostname_checker.match(name):
-        raise ValueError('endpoint name {} should be in http(s)://<hostname>[:<port>] and hostname may consist only of:'
-                         ' a-z, A-Z, 0-9, underscore and hyphens.'.format(name))
+        raise ValueError('endpoint name {} should be in http(s)://<hostname>'
+                         '[:<port>] and hostname may consist only of: '
+                         'a-z, A-Z, 0-9, underscore and hyphens.'.format(name))
 
 
 def _check_endpoint_name(name):
@@ -62,8 +64,8 @@ class Client(object):
         """
         Connects to a running server.
 
-        The class constructor takes a server address which is then used to connect
-        for all subsequent member APIs.
+        The class constructor takes a server address which is then used to
+        connect for all subsequent member APIs.
 
         Parameters
         ----------
@@ -249,7 +251,8 @@ class Client(object):
             description = 'Alias for %s' % existing_endpoint_name
 
         if existing_endpoint_name not in self.get_endpoints():
-            raise ValueError("Endpoint '%s' does not exist." % existing_endpoint_name)
+            raise ValueError(
+                "Endpoint '%s' does not exist." % existing_endpoint_name)
 
         # Can only overwrite existing alias
         existing_endpoint = self.get_endpoints().get(alias)
@@ -264,7 +267,8 @@ class Client(object):
 
         if existing_endpoint:
             if existing_endpoint.type != 'alias':
-                raise RuntimeError('Name "%s" is already in use by another endpoint.' % alias)
+                raise RuntimeError(
+                    'Name "%s" is already in use by another endpoint.' % alias)
 
             endpoint.version = existing_endpoint.version + 1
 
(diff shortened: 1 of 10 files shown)


tabpy-tools/tabpy_tools/client.py
```diff
@@ -21,7 +21,7 @@
 
 logger = logging.getLogger(__name__)
 
-_name_checker = compile('^[a-zA-Z0-9-_\ ]+$')
+_name_checker = compile(r'^[a-zA-Z0-9-_\ ]+$')
```
WillAyd: Haven't stepped through but I'm very surprised that this code would have worked without being a raw string

tabpy-tools/tabpy_tools/rest_client.py
```diff
@@ -146,21 +134,19 @@ def query(self, name, *args, **kwargs):
...
-                                'when querying an endpoint is not supported.')
-        return self.service_client.POST('query/'+name,
-            data={'data':args or kwargs},
-            timeout=self.query_timeout)
-
+            raise ValueError(
+                'Mixing of keyword arguments and positional arguments when '
+                'querying an endpoint is not supported.')
+        return self.service_client.POST('query/' + name,
+                                        data={'data': args or kwargs},
```
WillAyd: This looks kind of wonky; wouldn't be surprised if it fails when kwargs gets passed but orthogonal to change here

tabpy-tools/tabpy_tools/schema.py
```diff
@@ -42,20 +42,23 @@ def _generate_schema_from_example_and_description(input, description):
         # This should not fail unless there are bugs with either genson or
         # jsonschema.
         _validate(input, input_schema)
-    except Exception as e:
+    except Exception:
```
WillAyd: Generally want to avoid catching bare Exceptions. Especially here where it's not doing anything of value (a fuller traceback should be logged with appropriate configuration) but again orthogonal to change

tabpy-tools/tabpy_tools/schema.py
```diff
@@ -97,9 +100,11 @@ def generate_schema(input, output, input_description=None, output_description=No
                    'properties': {'y': {'type': 'integer', 'description': 'value of y'},
                                   'x': {'type': 'integer', 'description': 'value of x'}}},
          'output': {'type': 'integer', 'description': 'x times y'}}
-    '''
-    input_schema = _generate_schema_from_example_and_description(input, input_description)
-    output_schema = _generate_schema_from_example_and_description(output, output_description)
+    '''  # noqa: E501
```
WillAyd: Added noqa for now, but may want to look at something like the ipython directive which I think will actually render code for you

tabpy-tools/tabpy_tools/schema.py
```diff
@@ -97,9 +100,11 @@ def generate_schema(input, output, input_description=None, output_description=No
...
+    input_schema = _generate_schema_from_example_and_description(
+        input, input_description)
```
WillAyd: Using `input` as a variable shadows a built-in which is less than ideal; should be changed in a subsequent PR (typically appending an underscore works by convention)

tabpy-tools/tools_tests/test_client.py
```diff
@@ -8,14 +7,12 @@
...
 
-import requests
-
 
 class TestClient(unittest.TestCase):
 
     def setUp(self):
         self.client = Client("http://example.com/")
-        self.client._service = Mock()
+        self.client._service = Mock()  # TODO: should spec this
```
WillAyd: Because this isn't built to a predefined spec it makes very loose assertions. There's a typo in one of the tests that dangerously gets assigned as a property of the Mock instead of actually making an assertion

tabpy-tools/tools_tests/test_client.py
```diff
@@ -67,7 +64,8 @@ def test_query(self):
 
         self.assertEqual(self.client.query("foo", a=1, b=2, c=3), "ok")
 
-        self.client._service.query.sssert_called_once_with("foo", a=1, b=2, c=3)
+        self.client._service.query.assert_called_once_with(
```
WillAyd: See comment about about Mock. This was mistyped as `sssert_called_once_with`

tabpy-tools/tools_tests/test_rest.py
```diff
@@ -52,7 +52,7 @@ def test_GET_InvalidData(self):
         data = {'cat'}
         try:
             self.assertEqual(self.rnw.GET(url, data), 'json')
-        except:
+        except Exception:  # TODO: refactor this...
```
WillAyd: This is a rather strange test design to catch an Exception and make assertions about the type of exception raised. Would be better to use `self.assertRaises`

tabpy-tools/tabpy_tools/client.py
0golovatyi: `\ ` is not valid regex, shouldn't it be `\s`?

0golovatyi:
LGTM other then a couple of minor comments I added. Do you mind to open issues for the suspicious places of code you noticed?

tabpy-tools/tabpy_tools/client.py
WillAyd: Within the brackets it loses special meaning so should match literally on a forward slash

tabpy-tools/tabpy_tools/schema.py
WillAyd: Opened #212 for this

tabpy-tools/tabpy_tools/schema.py
WillAyd: See #213 for follow up

tabpy-tools/tools_tests/test_client.py
WillAyd: Follow up is #214

WillAyd:
Follow ups created

tabpy-tools/tabpy_tools/schema.py
WillAyd: What tool are you using for PEP8 compliance? Ran this with flake8 and no issues there so curious if I should be using something else

tabpy-tools/tools_tests/test_rest.py
0golovatyi: Agreed, we are slowly reworking tests as well

tabpy-tools/tabpy_tools/schema.py
0golovatyi: pep8speak

tabpy-tools/tabpy_tools/client.py
WillAyd: Hmm OK so looking more closely I think the original regex meant to use a character class of `\s` instead. I'll update on next PR

tabpy-tools/tabpy_tools/schema.py
WillAyd: Think we can get away with removing the exception block altogether? If the error propagates will write to stdout with a lot more information (specifically traceback)

WillAyd: Took a slightly different approach than what was suggested to use `logger.exception` which would print the stack trace as well.

Ran with pycodestyle 2.5.0 and came back clean. If you still get something on your end can you send me your pycodestyle version so I can try to reproduce?

PR #219, opened 2019-03-12
Fix readthedocs

0golovatyi:


docs/index.rst
```diff
@@ -1,31 +1,56 @@
...
+allows expanding Tableau with executing Python scripts on table
+calculation.
+
+
+ .. toctree::
+   :maxdepth: 2
+   :caption: General documentation:
+
+-  `About TabPy <about.md>`__
+-  `TabPy Server Download Instructions <server-download.md>`__
```
WillAyd: Should only need one trailing underscore for these to render, though I think as long as the title of the target file matches the description you are looking for it will render that would automatically (and be preferable)

docs/index.rst
```diff
@@ -1,31 +1,56 @@
...
+-  `TabPy Server Startup Instructions <server-startup.md>`__
+-  `Supported TabPy Server
+   Configurations <server-configurations.md>`__
+-  `Running TabPy in Python Virtual
+   Environment <tabpy-virtualenv.md>`__
+-  `Authoring Python calculations in
+   Tableau <TableauConfiguration.md>`__.
+-  `TabPy Tools <tabpy-tools.md>`__
+
+ .. toctree::
```
WillAyd: So rather than having multiple toctree directives like this (not sure how well that is supposed) might want to consider just creating subfolders in docs to house the various topics. You can then link to an `index.rst` within each subfolder to get the same effect with better organization

PR #222, opened 2019-03-21
Fix startup.cmd and tabpy.py

0golovatyi:


startup.cmd
```diff
@@ -43,33 +46,31 @@ IF %ERRORLEVEL% NEQ 0 (
 
 REM Parse optional CLI arguments: config file
 ECHO Parsing parameters...
-SET STARTUP_CMD=python tabpy-server\tabpy.py
+SET PYTHONPATH=%TABPY_ROOT%\tabpy-server;%PATH%
```
johng42: Is this even used?

tabpy-server/tabpy_server/tabpy.py
```diff
@@ -1,967 +1,10 @@
...
-import multiprocessing
-from OpenSSL import crypto
-import os
-from pathlib import Path
-from re import compile as _compile
-import requests
-import shutil
-import simplejson
-import sys
 from tabpy_server import __version__
```
johng42: Is this file change related to the rest of the fix?  It's really hard to tell from this one change.

startup.cmd
```diff
@@ -2,18 +2,20 @@
 SETLOCAL
 
 
-REM Check for Python in PATH
+REM Set environment variables.
+SET TABPY_ROOT=%CD%
+SET INSTALL_LOG=%TABPY_ROOT%\tabpy-server\install.log
```
johng42: Is this ever used?

startup.cmd
0golovatyi: Used by Python when searching packages

startup.cmd
0golovatyi: Used to append to log file

PR #224, opened 2019-03-26
Fix models

0golovatyi:


startup.sh
```diff
@@ -64,12 +64,15 @@ fi
 # Start TabPy server
 echo
 echo Starting TabPy server...
+SAVE_PYTHONPATH=$PYTHONPATH
+export PYTHONPATH=${TABPY_ROOT}/tabpy-server:${TABPY_ROOT}/tabpy-tools:$PYTHONPATH
```
jnegara: nit: can you shadow the system PYTHONPATH variable like you did in the batch script?

PR #225, opened 2019-03-27
Pbkdf2

0golovatyi:
- Password hashing is done with PBKDF2
- Added utility to add/update accounts
- Updated documentation


docs/server-config.md
```diff
@@ -41,3 +41,57 @@ TABPY_KEY_FILE = C:/path/to/key/file.key
...
+
+Password file is a text file containing usernames and hashed passwords
+per line separated by space. For username only ASCII characters
+supported.
+
+There is `utils/user_management.py` utility to operate with
+accounts in the password file. Run `utils/user_management.py -h` to
+see how to use it.
+
+Note for any of the changes for password file TabPy needs to be restarted.
```
johng42: This would be better
After making any changes to the  password file, TabPy needs to be restarted.

tabpy-server/tabpy_server/app/app.py
```diff
@@ -285,48 +287,18 @@ def _validate_cert_key_state(msg, cert_valid, key_valid):
...
-                        logger.error(
-                            'Multiple entries for username {} '
-                            'in password file'.format(login))
-                        return False
-
-                    self.credentials[login] = row[1]
-                    logger.debug('Found username {}'.format(login))
-
-            if len(self.credentials) == 0:
+            succeeded, self.credentials = parse_pwd_file(
```
johng42: is "succeeded," needed here?  (I'm asking - doesn't look like it would be since it would be true always)

tabpy-server/tabpy_server/app/app.py
```diff
@@ -285,48 +287,18 @@ def _validate_cert_key_state(msg, cert_valid, key_valid):
             log_and_raise(err, RuntimeError)
 
     def _parse_pwd_file(self):
+        succeeded = True
         if ConfigParameters.TABPY_PWD_FILE not in self.settings:
             logger.info("Password file is not specified")
```
johng42: should succeeded be set to false if there is  no password file specified?

tabpy-server/tabpy_server/app/util.py
```diff
@@ -35,3 +37,58 @@ def validate_cert(cert_file_path):
...
+    Parameters
+    ----------
+    pwd_file_name : str
+        Passwords file name.
+
+    Returns
+    -------
+    succeeded : bool
+        True if specified file was parsed successfully.
+        False if there were ane issues with parsing specified file.
```
johng42: typo   ane == any

tabpy-server/tabpy_server/app/util.py
```diff
@@ -35,3 +37,58 @@ def validate_cert(cert_file_path):
...
+
+    credentials : dict
+        Credentials from the file. Empty if succeeded is False.
```
johng42: It looks like this may not be true if the first few entries are valid but the Nth entry is a dupe (for instance).  need to test this assumption

utils/user_management.py
```diff
@@ -0,0 +1,158 @@
...
+        choices=['add', 'update'],
+        help='Command to execute')
+    parser.add_argument(
+        '-u',
+        '--username',
+        help='Username to add to passwords file')
+    parser.add_argument(
+        '-f',
+        '--pwdfile',
+        help='Passwords file')
```
johng42: Should be "Fully qualified path to passwords file"

utils/user_management.py
```diff
@@ -0,0 +1,158 @@
...
+def generate_password(pwd_len=16):
+    # List of characters to generate password from.
+    # We want to avoid to use similarly looking pairs like
+    # (O, 0), (1, l), etc.
+    lower_case_letters = 'abcdefghijkmnpqrstuvwxyz'
+    upper_case_letters = 'ABCDEFGHIJKLMPQRSTUVWXYZ'
+    digits = '23456789'
+
+    # and for punctuation we want to exclude some characters
+    # like inverted coma which can be hard to find and/or
```
johng42: typo comma

utils/user_management.py
```diff
@@ -0,0 +1,158 @@
...
+    # type
+    punctuation = '!#$%&()*+,-./:;<=>?@[\\]^_{|}~'
+
+    # we also want to have more letters and digits in
+    # generated password then punctuations
```
johng42: typo  then == than

utils/user_management.py
```diff
@@ -0,0 +1,158 @@
...
+    return True
+
+
+def generate_password(pwd_len=16):
+    # List of characters to generate password from.
+    # We want to avoid to use similarly looking pairs like
+    # (O, 0), (1, l), etc.
+    lower_case_letters = 'abcdefghijkmnpqrstuvwxyz'
+    upper_case_letters = 'ABCDEFGHIJKLMPQRSTUVWXYZ'
+    digits = '23456789'
```
johng42: why are 0 and 1 excluded if there is a random include being used (below)

docs/server-config.md
```diff
@@ -41,3 +41,57 @@ TABPY_KEY_FILE = C:/path/to/key/file.key
...
+## Authentication
+
+TabPy supports basic access authentication (see
+[https://en.wikipedia.org/wiki/Basic_access_authentication](https://en.wikipedia.org/wiki/Basic_access_authentication)
+for more details).
+
+### Enabling Authentication
+
+To enable the feature specify `TABPY_PWD_FILE` parameter in TabPy
+configuration file:
```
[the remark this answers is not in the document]
0golovatyi: Fixed

docs/server-config.md
0golovatyi: fixed

tabpy-server/tabpy_server/app/app.py
0golovatyi: parsing may fail due to many reasons, e.g. duplicated usernames, missing password hash, etc. Also, it may succeed and return empty credentials list for an empty file.

tabpy-server/tabpy_server/app/app.py
0golovatyi: This function is called without checking if the parameter was set in config or not and returns the result of parsing. For the case when there's no password file we don't want to fail.

But you are right - this code here is confusing, I'll rework it.

tabpy-server/tabpy_server/app/util.py
0golovatyi: fix

tabpy-server/tabpy_server/app/util.py
0golovatyi: Line 89 is `return False, {}` where `{}` is an empty credentials collection

utils/user_management.py
0golovatyi: fixed

utils/user_management.py
0golovatyi: fixed

utils/user_management.py
0golovatyi: fixed

utils/user_management.py
0golovatyi: Because 0 can be confused with O and 1 can be confused with l.

utils/user_management.py
```diff
@@ -0,0 +1,158 @@
...
+    return
+
+
+if __name__ == '__main__':
+    logging.basicConfig(level=logging.DEBUG, format="%(message)s")
+
+    # add tabpy-tools and tabpy-server folders to
+    # PYTHONPATH so code from there can be found when
+    # modules are imported
+    for dir_ in {'tabpy-tools', 'tabpy-server'}:
```
WillAyd: Is there a reason for this? Generally modifying the path like this from within code makes it generally more difficult to debug. 

At the very least I don't think you need to add the tools package here as you aren't importing from it

utils/user_management.py
```diff
@@ -0,0 +1,158 @@
...
+    return store_passwords_file(args.pwdfile, credentials)
+
+
+def process_command(args, credentials):
+    if args.command == 'add':
+        return add_user(args, credentials)
+    elif args.command == 'update':
+        return update_user(args, credentials)
+    else:
+        logger.error('Uknown command "%s"' % args.command)
```
WillAyd: Probably best to just use string formatting at this point instead of Py2 syntax

tabpy-server/tabpy_server/app/util.py
```diff
@@ -35,3 +37,58 @@ def validate_cert(cert_file_path):
...
+    '''
+    logger.info('Parsing passwords file %s...' % pwd_file_name)
```
WillAyd: Same thing on str formatting

tabpy-server/server_tests/test_evaluation_plane_handler.py
```diff
@@ -25,8 +25,11 @@ def setUpClass(cls):
         # create password file
         cls.pwd_file = tempfile.NamedTemporaryFile(
             mode='w+t', prefix=prefix, suffix='.txt', delete=False)
-        cls.pwd_file.write('username {}'.format(
-            hashlib.sha3_256('password'.encode('utf-8')).hexdigest()))
+        username = 'username'
+        password = 'password'
+        cls.pwd_file.write('{} {}\n'.format(
+            username,
+            hash_password(username, 'password')))
```
WillAyd: Guess it doesn't matter a lot here but you probably mean to reference the local variable rather than literal string in this expression

utils/user_management.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def check_args(args):
+    if (args.username is None) or (args.pwdfile is None):
+        return False
+
+    return True
+
+
+def generate_password(pwd_len=16):
```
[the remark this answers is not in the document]
0golovatyi: I checked https://docs.python.org/3/library/secrets.html and it shows how to generate a password in the same way we do in this function

utils/user_management.py
0golovatyi: The main reason we want for users to be able to run the tool from repo root folder and for that we need to have at least `tabpy-server` in `PYTHONPATH` as we reuse some code from it. As for `tabpy-tools` it is not needed now but may be needed in future - server code still uses some types from tools package :(

utils/user_management.py
0golovatyi: Ah, right. Somehow I keep forgetting about the recommended way for string formatting. Thanks for catching these!

tabpy-server/tabpy_server/app/util.py
0golovatyi: fixed

tabpy-server/server_tests/test_evaluation_plane_handler.py
0golovatyi: TY! fixed.

0golovatyi:
@WillAyd Answering your question about how secure is the password file. It is not, and we are aware of it. We are discussing some future improvements where we engage real authentication service instead of implementing our own. What we have here is more for auditing (so TabPy can log what account requested what operation).

However we considered security questions you raised and for now this is what our answer is:
- Using PBKDF2 for hashing makes it significantly harder to recover a password comparing with SHA (times if not powers).
- For the password file, we believe hosting OS mechanisms should be used to restrict access to files with sensitive information. I'll update the documentation with the recommendation.

utils/user_management.py
WillAyd: Gotcha. Not super important for now since this is just a util file but hoping we can effectively decouple the tools and server packages. Otherwise we should just bundle them together in code base

utils/user_management.py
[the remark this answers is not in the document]
WillAyd: Yea I was just thinking you could use `token_safe` unless you really cared to control the contents here. Simplifies code

tabpy-server/tabpy_server/app/util.py
```diff
@@ -35,3 +37,58 @@ def validate_cert(cert_file_path):
...
+            if login in credentials:
+                logger.error(
+                    'Multiple entries for username {} '
+                    'in password file'.format(login))
+                return False, {}
+
+            credentials[login] = row[1]
+            logger.debug('Found username {}'.format(login))
+
+    return True, credentials
```
WillAyd: So back to previous conversation around security, I think the README updates you made are great! But the problem is that regardless of OS permissions a user could technically still deploy a function to import and execute this function to return back the credentials of all users so long as they can make a reasonable guess at the file name (which probably wouldn't be that hard).

I think it would really help security a lot if credentials were never returned here nor exposable through anything that exists in the global namespace of any module

utils/user_management.py
WillAyd: Yep. By the way since you are on 3.6 you could simplify even further with formatted string literals, so just:

```python
f'Unknown command {args.command}'
```

Don't need to do it here but there's an open issue about it

utils/user_management.py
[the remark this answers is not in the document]
0golovatyi: If I read the documentation correctly `token_safe` returns base64 encoded token - there won't be punctuation characters in it.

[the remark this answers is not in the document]
WillAyd: There is URL-safe punctuation. You can see the PEP for the module here:

https://www.python.org/dev/peps/pep-0506/#id42

Basically the whole point of the module is to prevent people from using `random` to generate passwords, which is what we are doing here. Random was just kept for backwards compatibility

utils/user_management.py
0golovatyi: Oh, believe me, we are constantly discussing how we can setup process more robust and platform independent.
There is a list of options but neither of them is going to work smoothly for all of the scenarios people use TabPy with.

utils/user_management.py
0golovatyi: Great, will try to remember that.

tabpy-server/tabpy_server/app/util.py
0golovatyi: Yeah, it's an architectural problem caused by TabPy executing user scripts in the same context where the Web server runs. 

It is not that bad as we are dealing with hashes of passwords and TabPy never knows actual passwords. Another limiting factor for an attacker is he/she needs to know at least one pair of login/password to be able to evaluate a script. And since all the users are the same to TabPy as it is implemented now (we don't have any authorization or resources restrictions) knowing additional login/password won't give an attacker any benefit.

WillAyd: Yea for sure. So my point is that it's just pretty easy for someone to get a hold of these hashed passwords, and that makes things difficult but obviously not impossible to target a particular user (say CEO of company) and try to match the hash with their own computing resources.

It would be a much better design if there was a database locked with a password available only to the deployed server which stored passwords and would be checked against per user on demand rather than having them persist in the memory space of Python

utils/user_management.py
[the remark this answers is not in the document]
0golovatyi: Got it (I think), pushing updated code.

tabpy-server/tabpy_server/app/util.py
0golovatyi: Agreed, but it also could be configuration nightmare if we require people to configure a DB and access to it considering TabPy setup itself can be challenging for some platforms.

Again, even with using basic access authentication here we see it more like auditing mechanism at the moment.

WillAyd: Yea makes sense. SQLite would be pretty easy to leverage here for future reference

Agreed it's an improvement over what's in place regardless

PR #226, opened 2019-03-28
Assorted cleanups from Py2

WillAyd:
Started with the intention of just cleaning up Py2 legacy stuff but threw in some other assorted cleanups


tabpy-tools/tabpy_tools/rest.py
```diff
@@ -214,7 +214,7 @@ def __delete__(self, instance):
         delattr(instance, self.name)
 
 
-class _RESTMetaclass(type(_MutableMapping)):
+class _RESTMetaclass(abc.ABCMeta):
```
WillAyd: Not sure the intention of this class but most metaclasses would define __new__ instead of __init__. Actually not sure the intentions of the latter but makes it seem like more of a base class than a meta

0golovatyi:
@WillAyd there are build failures caused by your change - https://travis-ci.com/tableau/TabPy/builds/106264800

WillAyd:
@0golovatyi thanks for the ping. Missing import which should be fixed now

PR #229, opened 2019-03-28
add required packages to setup.py

johng42:
Many of these packages are included in a typical base install of Python.  Testing has revealed that different linux distributions will customize python installs and omit one or more of these packages.  Add all known discrepancies uncovered during testing to setup.py and add a comment to this effect.

### DIFF
--- a/tabpy-server/setup.py
+++ b/tabpy-server/setup.py
@@ -28,13 +28,27 @@ setup(
     package_data={'tabpy_server.static': ['*.*'],
                   'tabpy_server': ['startup.*', 'state.ini']},
     license='MIT',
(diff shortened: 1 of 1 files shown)


tabpy-server/setup.py
```diff
@@ -28,13 +28,27 @@
     package_data={'tabpy_server.static': ['*.*'],
                   'tabpy_server': ['startup.*', 'state.ini']},
     license='MIT',
+# Note: many of these required packages are included in base python
```
0golovatyi: Comments should be aligned with code according to pep8

johng42: fixed

PR #230, opened 2019-03-28
Dev flake8 server

0golovatyi:
Copying changes from master made by @WillAyd.

Lesson learned - do not `git cherry-pick` unless you want to end up with merging conflicts which are harder to solve then manually duplicate all necessary changes.


WillAyd:
> Lesson learned - do not `git cherry-pick` unless you want to end up with merging conflicts which are harder to solve then manually duplicate all necessary changes.

Ha whoops sorry about that. Typically should work well but if you want to try again in the future let me know can see if I can help

johng42:
CC drop is expected since some single lines are broken into multiple lines by PEP changes

PR #233, opened 2019-03-29
Fix restricted evaluation bug

0golovatyi:
The bug was caused by code change where we started treating each entry in config file as a string instead of parsing port as int and so on.

### DIFF
--- a/tabpy-server/tabpy_server/common/endpoint_file_mgr.py
+++ b/tabpy-server/tabpy_server/common/endpoint_file_mgr.py
@@ -15,7 +15,7 @@ from re import compile as _compile
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


tabpy-server/tabpy_server/common/endpoint_file_mgr.py
```diff
@@ -15,7 +15,7 @@
 
 logger = logging.getLogger(__name__)
 
-_name_checker = _compile(r'^[a-zA-Z0-9-_\ ]+$')
+_name_checker = _compile('^[a-zA-Z0-9-_\ ]+$')
```
WillAyd: I think you just want to use the built in character class of `\w` for this

0golovatyi: Yes, we'll have another fix for regex. With `r` prefix models deployment doesn't work so need to look into it.

PR #238, opened 2019-04-01
Add CI version update script

0golovatyi:
With this CI script we can have commit as a part of TabPy version


0golovatyi:
For this to work we need to set up github API key for Travis as described at https://stackoverflow.com/questions/23277391/how-to-publish-to-github-pages-from-travis-ci.

@jnegara could you take a look?

PR #244, opened 2019-04-04
update pw file parser

johng42:
If the password file had a username followed by a space, parsing would succeed and set the user password to null
add a check to fail parsing and log a warning about this conditin


0golovatyi:
Great catch! Can we have a unit test for that?

jnegara:
Why isn't coverage reflecting the new test?

PR #245, opened 2019-04-04
WIP: Add predeployed models to TabPy

sbabayan:
Updated documentation, added an integration test, and implemented PCA/SentimentAnalysis


.travis.yml
```diff
@@ -10,6 +10,7 @@ install:
 script:
    - pytest tabpy-server/server_tests/ --cov=tabpy-server/tabpy_server
    - pytest tabpy-tools/tools_tests/ --cov=tabpy-tools/tabpy_tools --cov-append
+   - pytest tests/integration_tests/ --cov=models --cov-append
```
jnegara: Can you update ```CONTRIBUTING.md``` to include instructions for running the integration tests?

models/scripts/SentimentAnalysis.py
```diff
@@ -0,0 +1,35 @@
...
+    elif libary == 'textblob':
+        for text in _arg1:
+            currScore = TextBlob(text)
+            scores.append(currScore.sentiment.polarity)
+    return scores
+
+
+if __name__ == '__main__':
+    port = sys.argv[1]
+    connection = Client('http://localhost:' + port + '/')
```
jnegara: These will need to be updated once tools supports auth.

docs/tabpy-tools.md
```diff
@@ -136,6 +137,95 @@ client.remove('WillItDefault')
 
 ```
 
+## Predeployed Functions
+
```
0golovatyi: There's build failure with markdownlint finding issues with your documentation, see here- https://travis-ci.com/tableau/TabPy/builds/107109046

docs/tabpy-tools.md
```diff
@@ -136,6 +137,95 @@ client.remove('WillItDefault')
 
 ```
 
+## Predeployed Functions
```
0golovatyi: @nmannheimer Are you proofreading these documentation updates?

models/scripts/PCA.py
```diff
@@ -0,0 +1,55 @@
+from tabpy_tools.client import Client
```
0golovatyi: I suggest we have docstring for all the new file. In this case, we can explain what is this model and where is the documentation for how to use it.

docs/tabpy-tools.md
```diff
@@ -136,6 +137,95 @@ client.remove('WillItDefault')
...
+```
+
+#### Sentiment Analysis
+
+[Sentiment analysis](https://en.wikipedia.org/wiki/Sentiment_analysis) is a technique
+which uses natural language processing to extract the emotional positivity or negativity
+– the sentiment – behind a piece of text and converts that into a numeric value. Our
+implementation of sentiment analysis returns a polarity score between -1 and 1 which rates
+the positivity of the string with 1 being very positive and -1 being very negative. Calling
+the Sentiment Analysis function from TabPy in Tableau will look like the following, where _arg1
```
0golovatyi: FYI all underscore characters to not make test italic need to be prefixed with \

docs/tabpy-tools.md
```diff
@@ -136,6 +137,95 @@ client.remove('WillItDefault')
...
+
+```python
+
+python setup.py 4047 
+
+```
+
+The setup file will install all of the necessary dependencies (sklearn, nltk,
+textblob, pandas, & numpy) and deploy all of the prebuilt models located in
+./models/scripts. For every model that is successfully deployed a message will
```
0golovatyi: For folders and file names same as function and package names I suggest we surround them with ` quotes.

models/scripts/PCA.py
```diff
@@ -0,0 +1,55 @@
...
+    pcaComponents = pca.fit_transform(scaledData)
+
+    return pcaComponents[:, component - 1].tolist()
+
+
+if __name__ == '__main__':
+    port = sys.argv[1]
+    connection = Client('http://localhost:' + port + '/')
+    connection.deploy('PCA', PCA, 'Returns the specified principal component.', override=True)
+    print("Successfully deployed PCA")
```
0golovatyi: How do you know it is successfully deployed?

models/scripts/SentimentAnalysis.py
0golovatyi: +1

tabpy-server/tabpy_server/state.ini
```diff
@@ -11,5 +11,4 @@ Access-Control-Allow-Methods =
 [Query Objects Docstrings]
 
 [Meta]
-Revision Number = 1
-
+Revision Number = 1
```
0golovatyi: To restore `state.ini` or any other file reverting all the changes run `git checkout HEAD -- tabpy-server/tabpy_server/stati.ini`

models/setup.py
```diff
@@ -0,0 +1,39 @@
+import pip
```
0golovatyi: Is it documented how to run integration tests locally?

models/scripts/SentimentAnalysis.py
sbabayan: added a to do comment to both PCA and SentimentAnalysis

models/setup.py
sbabayan: i added it to CONTRIBUTING.md

models/scripts/PCA.py
sbabayan: connectin.deploy throws an error if it isn't able to deploy the function so the print statement can only get hit if the function successfully deploys

docs/tabpy-tools.md
sbabayan: @nmannheimer can you please review the docstrings I added to PCA and SentimentAnalysis as well

nmannheimer: Yep! Reviewing now.

CONTRIBUTING.md
```diff
@@ -61,11 +61,12 @@ install `pytest`:
 pip install pytest
 ```
 
-And then run `pytest` either for server or tools test, or even combined:
+And then run `pytest` either for server, tools, or integrations tests, or even combined:
 
 ```sh
 pytest tabpy-server/server_tests/ --cov=tabpy-server/tabpy_server
 pytest tabpy-tools/tools_tests/ --cov=tabpy-tools/tabpy_tools --cov-append
+pytest tabpy-tools/tests/integration_tests/ --cov=models --cov-append
```
jnegara: Might want to note that the integration tests have side effects.

0golovatyi: @sbabayan I think instead of documenting it and make people to revert `state.ini` after every local test run the tests should point TabPy to a temporary `state.ini` in the same way we do for config and passwords files in unit tests.

@jnegara Great catch, we all overlooked this.

jnegara: @0golovatyi Even if we implement what you propose, the test would still install Python packages that are not necessary for running vanilla TabPy to the machine on which it is running. This should be documented.

0golovatyi: Agree. Packages we add to Python env is unrelated issue IMO and needs to be documented per model.

docs/tabpy-tools.md
```diff
@@ -136,6 +137,102 @@ client.remove('WillItDefault')
 
 ```
 
+## Predeployed Functions
+
+To setup models download the latest version of TabPy and follow the instructions
```
nmannheimer: maybe link to the install instructions?

docs/tabpy-tools.md
```diff
@@ -136,6 +137,102 @@ client.remove('WillItDefault')
...
+to install and start up your server. Once your server is running, navigate to the
+models directory and run setup.py.  If your TabPy server is running on the default
+port (9004), you do not need to specify a port when launching the script. If your
+server is running on a port other than 9004 you can specify a different port in
+the command line like so
```
nmannheimer: just a formatting nit, I would add : at the end of these

docs/tabpy-tools.md
```diff
@@ -136,6 +137,102 @@ client.remove('WillItDefault')
...
+```sh
+
+python setup.py 4047
+
+```
+
+The setup file will install all of the necessary dependencies `(sklearn, nltk,
+textblob, pandas, & numpy)` and deploy all of the prebuilt models located in
+`./models/scripts`. For every model that is successfully deployed a message
+will be printed to the console
```
nmannheimer: :

docs/tabpy-tools.md
```diff
@@ -136,6 +137,102 @@ client.remove('WillItDefault')
...
+is a statistical technique which extracts new, linearly uncorrelated,
+variables out of a dataset which capture the maximum variance in the
+data. In this way, `PCA` can be used to reduce the number of variables
+in a high dimensional dataset, a process that is called dimensionality
+reduction. The first principal component captures the largest amount of
+variance, while the second captures the largest portion of the remaining
+variance while remaining orthogonal to the first and so on. This allows the
+reduction of the number of dimensions while maintaining as much of the
+information from the original data as possible. `PCA` is useful in
+exploratory data analysis because complex relationships can be visualized
```
nmannheimer: "complex **linear** relationships"

docs/tabpy-tools.md
```diff
@@ -136,6 +137,102 @@ client.remove('WillItDefault')
...
+[decomposition.PCA](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html)
+algorithm, which is further documented
+[here](https://scikit-learn.org/stable/modules/decomposition.html#pca).
+It requires the selected component to be > 0 and <= number of variables you
+pass in to the function. When passing categorical variables we perform the
+`scikit-learn` [One Hot Encoding](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OneHotEncoder.html)
+to transform your non-numeric variables into a table of 0s and 1s. In order
+for `One Hot Encoding` to be performant we have limited the number of unique
+values your categorical column may contain to 25 and do not permit any nulls
+or empty strings in the column. Before `PCA` is performed, all variables are
```
nmannheimer: In Tableau's implementation of  `PCA`, all variables are...

docs/tabpy-tools.md
```diff
@@ -136,6 +137,102 @@ client.remove('WillItDefault')
...
+in a 2D scatter plot of the first few principal components.
+
+TabPy’s implementation of `PCA` uses the scikit-learn
+[decomposition.PCA](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html)
+algorithm, which is further documented
+[here](https://scikit-learn.org/stable/modules/decomposition.html#pca).
+It requires the selected component to be > 0 and <= number of variables you
+pass in to the function. When passing categorical variables we perform the
+`scikit-learn` [One Hot Encoding](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OneHotEncoder.html)
+to transform your non-numeric variables into a table of 0s and 1s. In order
```
nmannheimer: variables into a one-hot numeric array of 0s and 1s.

docs/tabpy-tools.md
```diff
@@ -136,6 +137,102 @@ client.remove('WillItDefault')
...
+for `One Hot Encoding` to be performant we have limited the number of unique
+values your categorical column may contain to 25 and do not permit any nulls
+or empty strings in the column. Before `PCA` is performed, all variables are
+normalized to have a mean of 0 and unit variance using the `scikit-learn`
+[StandardScaler](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html).
+
+A Tableau calculated field to perform PCA will look like
```
nmannheimer: :

docs/tabpy-tools.md
```diff
@@ -136,6 +137,102 @@ client.remove('WillItDefault')
...
+reduction of the number of dimensions while maintaining as much of the
+information from the original data as possible. `PCA` is useful in
+exploratory data analysis because complex relationships can be visualized
+in a 2D scatter plot of the first few principal components.
+
+TabPy’s implementation of `PCA` uses the scikit-learn
+[decomposition.PCA](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html)
+algorithm, which is further documented
+[here](https://scikit-learn.org/stable/modules/decomposition.html#pca).
+It requires the selected component to be > 0 and <= number of variables you
```
nmannheimer: Say something like:

In the script, after the function name 'PCA', a specific principal component to return must be selected. This integer input should be > 0 and <= number of variables you pass into the function.

docs/tabpy-tools.md
```diff
@@ -136,6 +137,102 @@ client.remove('WillItDefault')
...
+### Sentiment Analysis
+
+[Sentiment analysis](https://en.wikipedia.org/wiki/Sentiment_analysis) is
+a technique which uses natural language processing to extract the emotional
+positivity or negativity – the sentiment – behind a piece of text and converts
+that into a numeric value. Our implementation of `sentiment analysis` returns a
+polarity score between -1 and 1 which rates the positivity of the string with
+1 being very positive and -1 being very negative. Calling the `Sentiment
+Analysis` function from TabPy in Tableau will look like the following,
+where \_arg1 is a dimension containing text
```
nmannheimer: is a Tableau dimensions containing

CHANGELOG
```diff
@@ -3,6 +3,11 @@
 This file list notable changes for TabPy project releases.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## v1.0
```
0golovatyi: Wait, wait, wait! We haven't released 0.4 yet :D

CHANGELOG
```diff
@@ -3,6 +3,11 @@
...
+
+### Improvements
+- Added a setup.py file to build and deploy PCA & Sentiment Analysis as pickled files
```
0golovatyi: Maybe something like "scripts and documentation for models"?

CHANGELOG
```diff
@@ -3,10 +3,10 @@
 This file list notable changes for TabPy project releases.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## v1.0
+## v0.5
```
0golovatyi: You'll need to change VERSION file with this.

models/setup.py
```diff
@@ -0,0 +1,40 @@
...
+def install_dependencies(packages):
+    pip_arg = ['install'] + packages + ['--no-cache-dir']
+    if hasattr(pip, 'main'):
+        pip.main(pip_arg)
+    else:
+        pip._internal.main(pip_arg)
+
+
+if __name__ == '__main__':
+    install_dependencies(['sklearn', 'pandas', 'numpy',
```
WillAyd: So I'd generally advise against having multiple setup files like this within your application. It is entirely possible at certain points in times that dependencies here may conflict with the overall application dependencies, so doing it in this fashion could upgrade / downgrade packages in a way that's not intuitive to end users.

You might alternately want to consider using a requirements file for conda which has optional dependencies and make sure that users install optional ones if they want these models. It would be a rather large scope and probably worth a separate PR from this but I think nonetheless would make things easier

models/setup.py
```diff
@@ -0,0 +1,40 @@
...
+                          'textblob', 'nltk', 'scipy'])
+    # Determine if we run python or python3
+    if platform.system() == 'Windows':
```
WillAyd: Generally I would advise against doing this; I haven't come across other packages where the application tries to manage this for you. Users typically need to make sure that python is an alias for python3 if they have multiple installations

models/setup.py
```diff
@@ -0,0 +1,40 @@
...
+    else:
+        py = 'python3 '
+
+    # When no port is specified we will assume the default of 9004
+    if len(sys.argv) == 1:
+        port = 9004
+    else:
+        port = sys.argv[1]
+    # Deploy each model in the scripts directory
+    directory = os.path.join(os.path.dirname(os.path.abspath(__file__)),
```
WillAyd: Might consider using pathlib instead of os.path to make code easier

models/setup.py
```diff
@@ -0,0 +1,40 @@
...
+                             'scripts/')
+    for filename in os.listdir(directory):
+        path = py + directory + filename + ' ' + str(port)
+        os.system(path)
```
WillAyd: Not sure what this is actually supposed to do (execute some kind of command?) but in any case it's not really advised to use os.system. See https://docs.python.org/3.7/library/subprocess.html#replacing-os-system

tests/integration_tests/test_deploy_model.py
```diff
@@ -0,0 +1,88 @@
...
+import time
+import platform
+import signal
+import tempfile
+import shutil
+
+
+class TestDeployModel(unittest.TestCase):
+    def __init__(self, *args, **kwargs):
+        super(TestDeployModel, self).__init__(*args, **kwargs)
```
WillAyd: You don't need this syntax in Python3 - can simplify as `super().__init__(*args, **kwargs)`

tests/integration_tests/test_deploy_model.py
```diff
@@ -0,0 +1,88 @@
...
+        self.tabpy_root = os.path.join(os.path.dirname(os.path.abspath(__file__)), '..', '..')
+        if platform.system() == 'Windows':
+            self.py = 'python'
+        else:
+            self.py = 'python3'
+
+    def setUp(self):
+        prefix = '_TestDeployModel_'
+        self.state_dir = tempfile.mkdtemp(prefix=prefix, dir=self.cwd)
+        self.state_file = open(os.path.join(self.state_dir, 'state.ini'), 'w+')
```
WillAyd: I'd always advise using a context manager instead of explicitly trying to do open...close with a file handle.

tests/integration_tests/test_deploy_model.py
```diff
@@ -0,0 +1,88 @@
...
+        os.chdir(self.tabpy_root)
+        # start TabPy server in the background
+        if platform.system() == 'Windows':
+            self.process = subprocess.Popen(['startup.cmd', os.path.basename(self.config_file.name), '&'])
+        else:
+            self.process = subprocess.Popen(['./startup.sh', '--config=' + os.path.basename(self.config_file.name), '&'],
+                                             preexec_fn=os.setsid)
+        time.sleep(1)
+
+    def tearDown(self):
```
WillAyd: Do you want to use the classmethod here? Seems pretty expensive to kill the TabPy server on every test instance

WillAyd: Note that this is equivalent for now since there is only one test, but if you add more tests killing the serve each time seems like a lot

tests/integration_tests/test_deploy_model.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def test_deploy_model_with_script(self):
+        """
+        Deploys a model using the provided deployment script.
+        
+        :return:
+        """
+        # run script
+        path = os.path.join(self.tabpy_root, 'models', 'setup.py')
+        os.system(self.py + ' ' + path)
```
WillAyd: Similar comment on using os.system

tests/integration_tests/test_deploy_model.py
```diff
@@ -0,0 +1,88 @@
...
+        self.process.kill()
+
+    def test_deploy_model_with_script(self):
+        """
+        Deploys a model using the provided deployment script.
+        
+        :return:
+        """
+        # run script
+        path = os.path.join(self.tabpy_root, 'models', 'setup.py')
```
WillAyd: Similar comment on using pathlib instead of os.path

models/scripts/PCA.py
```diff
@@ -0,0 +1,66 @@
...
+    to capture as much of the variance as possible. For more information
+    on the function and how to use it please refer to tabpy-tools.md
+    '''
+    cols = [_arg1, _arg2] + list(_argN)
+    encodedCols = []
+    labelEncoder = LabelEncoder()
+    oneHotEncoder = OneHotEncoder(categories='auto', sparse=False)
+
+    for col in cols:
+        if (type(col[0]) is int or type(col[0]) is float):
```
WillAyd: More idiomatic to write this as `isinstance(col[0], (int, float))`

models/scripts/PCA.py
```diff
@@ -0,0 +1,66 @@
...
+                assert (False)
+            integerEncoded = labelEncoder.fit_transform(array(col))
+            integerEncoded = integerEncoded.reshape(len(col), 1)
+            oneHotEncoded = oneHotEncoder.fit_transform(integerEncoded)
+            transformedMatrix = oneHotEncoded.transpose()
+            encodedCols += list(transformedMatrix)
+
+    dataDict = {}
+    for i in range(len(encodedCols)):
+        dataDict['col' + str(1 + i)] = list(encodedCols[i])
```
WillAyd: Can just use f-strings here, so `f'col{1 + i}'`

tests/integration_tests/test_deploy_model.py
sbabayan: fixed

models/setup.py
sbabayan: fixed

tests/integration_tests/test_deploy_model.py
sbabayan: fixed

models/scripts/PCA.py
sbabayan: fixed

models/scripts/PCA.py
sbabayan: thanks fixed

models/setup.py
sbabayan: fixed

tests/integration_tests/test_deploy_model.py
sbabayan: fixed

models/setup.py
sbabayan: Yeah you're completely right and it's definitely a known risk. However, since these pre-deployed models are an optional feature and the packages aren't installed by default we chose to implement them this way for the time being.

models/setup.py
sbabayan: I'm not sure that would work for anaconda or jupyter? We're hoping to make the process as simple as possible for users who choose to "opt-in" since we aren't installing them by default.

tests/integration_tests/test_deploy_model.py
sbabayan: Since all of the other unit tests use open/close I chose to follow the same pattern. However I think we have plans to revamp our tests in the near future so hopefully we can address it then.

tests/integration_tests/test_deploy_model.py
sbabayan: Thanks for all of your feedback, it's been very helpful especially in regards to python3. 

We actually want to start and stop a server for every test since we don't want any failures related to models to affect the test runs for other features. We also have somewhat specific tabpy environments for the tests we're in the process of adding so it wouldn't be possible to use the same server for each of those tests.

models/setup.py
WillAyd: What distinction are you trying to make between platforms here? This is really only going to cause confusion or do nothing

0golovatyi: I am not saying this is the right solution we have here (didn't have enough time to carefully look at all the changes) but just want to note that in Anaconda there's no python3 even for Python 3.x environment - there's just python.
As for making users to create aliases for python3... we need to think more about it. We'll need steps for Linux, mac, windows, anaconda, Jupiter and who knows what else. Even now we can see people ignore documented installation steps and after downloading and unpacking zip file trying to be creative for how to configure and start TabPy :)

WillAyd: So still not totally clear on the problem being solved here but I would say that by trying to manage this in TabPy you'd be signing up for a lot of non-value added activities. In a very simplistic case, if a user had `python` and `python3` pointing to different environments you'd start executing different interpreters by doing something like this. Even worse, `pip` and `pip3` would probably be two different aliases so you'd be installing packages potentially into a different version of Python than you ultimately want to execute; this will be very confusing to diagnose and help users with.

It's not going to be done in the PR but again just giving a very strong +1 towards providing wheel distributions of TabPy. Requiring a source build and attempting to manually manage environments for people within the package is not something I am aware of any other Python package requiring

tests/integration_tests/test_deploy_model.py
WillAyd: Gotcha. So what I would suggest is that every environment you want to configure would be its own class with class methods to start up and tear down the server appropriately. That would keep things organized and get you the same effect you want. By starting and stopping every test we wouldn't really be mirroring production usage, since it's unrealistic for end users to do the same thing in a production environment after every endpoint gets queried.

Again realize that this is easier said than done, so not something that needs to be done in this PR but should be given consideration

docs/tabpy-tools.md
```diff
@@ -8,6 +8,7 @@ on TabPy server.
 - [Connecting to TabPy](#connecting-to-tabpy)
 - [Authentication](#authentication)
 - [Deploying a Function](#deploying-a-function)
+- [Predeployed Functions](#predeployed-functions)
```
0golovatyi: TOC is updated with markdown-toc as described in CONTRIBUTING.md. You shouldn't edit it manually. Also, individual models are not listed in here.

docs/tabpy-tools.md
```diff
@@ -157,6 +158,114 @@ client.remove('WillItDefault')
...
+script. If your server is running using a custom config you can specify the config
+in the command line like so:
+
+```sh
+
+python setup.py custom.conf
+
+```
+
+The setup file will install all of the necessary dependencies `(sklearn, nltk,
```
0golovatyi: I would say "e.g. sklearn, nltk, pandas"

models/scripts/PCA.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+if __name__ == '__main__':
+    #running from setup.py
+    if (len(sys.argv) > 1):
+        config_file_path = sys.argv[1]
+    else: 
+        config_file_path = str(Path(__file__).resolve().parent.parent.parent
+                               / 'tabpy-server' / 'tabpy_server' / 'common'
+                               / 'default.conf')
```
0golovatyi: For investigation purposes, we need to log what config file is parsed, what password file we found if any and so on.

models/scripts/SentimentAnalysis.py
```diff
@@ -0,0 +1,69 @@
...
+    elif libary == 'textblob':
+        for text in _arg1:
+            currScore = TextBlob(text)
+            scores.append(currScore.sentiment.polarity)
+    return scores
+
+
+if __name__ == '__main__':
+    #running from setup.py
+    if (len(sys.argv) > 1):
```
0golovatyi: Lines 40-64 are duplicated. This is a common functionality which should live in its own function and be reused by all models.

docs/tabpy-tools.md
```diff
@@ -157,6 +158,114 @@ client.remove('WillItDefault')
...
+To setup models download the latest version of TabPy and follow the [instructions](server-download.md)
+to install and start up your server. Once your server is running, navigate to the
+models directory and run setup.py.  If your TabPy server is running on the default
+config (default.conf), you do not need to specify a config file when launching the
+script. If your server is running using a custom config you can specify the config
+in the command line like so:
+
+```sh
+
+python setup.py custom.conf
```
0golovatyi: What about login and password parameters?

models/scripts/PCA.py
sbabayan: This is logged in setup already via printing to the console

docs/tabpy-tools.md
sbabayan: fixed

docs/tabpy-tools.md
sbabayan: I removed my TOC edit. The individual models are listed below Predeployed Functions, they don't need to be listed specifically in the TOC especially if we continue to add more the list will grow very long

models/scripts/SentimentAnalysis.py
sbabayan: fixed

docs/tabpy-tools.md
sbabayan: users are prompted for their login/password if auth is turned on in their custom conf, they don't pass it as arguments to the setup file

0golovatyi:
Oh, one thing - could you update `VERSION` file?

models/scripts/SentimentAnalysis.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def SentimentAnalysis(_arg1, library='nltk'):
+    '''
+    Sentiment Analysis is a procedure that assigns a score from -1 to 1
+    for a piece of text with -1 being negative and 1 being positive. For
+    more information on the function and how to use it please refer to
+    tabpy-tools.md
+    '''
+    assert (type(_arg1[0]) is str)
```
WillAyd: Just do `isinstance(_arg1[0],. str)` instead of passing as argument to `type`

models/scripts/SentimentAnalysis.py
```diff
@@ -0,0 +1,59 @@
...
+
+    library = library.lower()
+    supportedLibraries = {'nltk', 'textblob'}
+
+    assert (library in supportedLibraries)
```
WillAyd: Just so you are aware assertion checks get stripped out if Python is run with `-OO` optimization, which some users may opt for. Typically better and more descriptive to raise some other type of error (here maybe a ValueError for the library)

models/setup.py
```diff
@@ -0,0 +1,50 @@
...
+    port, auth_on, prefix = setup_utils.parse_config(config_file_path)
+    if auth_on:
+        auth_args = setup_utils.get_creds()
+    else:
+        auth_args = []
+
+    directory = str(Path(__file__).resolve().parent / 'scripts')
+    # Deploy each model in the scripts directory
+    for filename in os.listdir(directory):
+        subprocess.call([py, f'{directory}/{filename}', config_file_path]
```
WillAyd: Call is part of the legacy Python API to support code prior to 3.5 - I think you'd be better served porting this over to `run` given 3.6 is the min supported version

models/utils/setup_utils.py
```diff
@@ -0,0 +1,30 @@
+import configparser
+from pathlib import Path
+import getpass
+import sys
+
+def get_default_config_file_path():
+    config_file_path = str(Path(__file__).resolve().parent.parent.parent
```
WillAyd: Minor nit but instead of 4 division operations could just do this once with the rhs operand being `'tabpy-server/tabpy_server/common/default.conf'`

tests/integration/integ_test_base.py
```diff
@@ -202,11 +202,13 @@ def setUp(self):
         # startup script
         with open(self.tmp_dir + '/output.txt', 'w') as outfile:
             if platform.system() == 'Windows':
+                self.py = 'python'
                 self.process = subprocess.Popen(
                     ['startup.cmd', self.config_file_name],
                     stdout=outfile,
                     stderr=outfile)
             else:
+                self.py = 'python3'
```
WillAyd: I still strongly advise against doing this type of aliasing; this is really only going to confuse people

models/setup.py
WillAyd: Another option would be to use runpy in combination with multiprocessing - that would ensure that whatever interpreter is running TabPy is the one that ultimately executes these scripts

tests/integration/integ_test_base.py
0golovatyi: I agree we need better way of finding out Python executable name on a platform. My suggestion is @sbabayan opens issue for doing so and we investigate (or search for how others solve the problem). IMO no need to delay these changes for the imperfection we see here.

models/scripts/SentimentAnalysis.py
sbabayan: I removed all of the asserts and raised errors instead

models/scripts/SentimentAnalysis.py
sbabayan: fixed

models/setup.py
sbabayan: thanks Will, I went ahead and updated it to be run I wasn't able to figure out a way to use runpy and specify command line arguments

models/utils/setup_utils.py
sbabayan: fixed, I'm not sure why I had it that way before thanks for catching that

tests/integration/integ_test_base.py
sbabayan: I will open an issue to track this work

PR #248, opened 2019-04-04
Unit tests for set_credentials in tabpy-tools.

jnegara:
Added unit tests for the new set_credentials methods in tabpy-tools.


tabpy-tools/tabpy_tools/rest.py
```diff
@@ -3,11 +3,11 @@
 import requests
 from requests.auth import HTTPBasicAuth
 from re import compile
+import json
```
0golovatyi: do we need both? simplejson in Python3 is just a wrapper for json for backward compatibility

jnegara: We reference both on line 27 & 28. I was just cleaning up line 9 from the old file.

Want me to refactor?

0golovatyi: Depends on how much work it is. Either refactor or open an issue (and TFS bug) for it.

jnegara: no need to open a bug, removed simplejson in the latest commit

PR #252, opened 2019-04-09
weren't setting the path properly for the optional config parameter

sbabayan:
Manually tested on both Mac & Windows creating a temporary config file and passing it in on startup

### DIFF
--- a/startup.cmd
+++ b/startup.cmd
@@ -50,7 +50,7 @@ SET PYTHONPATH=%TABPY_ROOT%\tabpy-server;%TABPY_ROOT%\tabpy-tools;%PYTHONPATH%
 SET STARTUP_CMD=python tabpy-server\tabpy_server\tabpy.py
 IF [%1] NEQ [] (
     ECHO     Using config file at %TABPY_ROOT%\tabpy-server\tabpy_server\%1
-    SET STARTUP_CMD=%STARTUP_CMD% --config=%1
+    SET STARTUP_CMD=%STARTUP_CMD% --config=%TABPY_ROOT%\tabpy-server\tabpy_server\%1
 )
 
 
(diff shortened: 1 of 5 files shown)


startup.cmd
```diff
@@ -50,7 +50,7 @@ SET PYTHONPATH=%TABPY_ROOT%\tabpy-server;%TABPY_ROOT%\tabpy-tools;%PYTHONPATH%
 SET STARTUP_CMD=python tabpy-server\tabpy_server\tabpy.py
 IF [%1] NEQ [] (
     ECHO     Using config file at %TABPY_ROOT%\tabpy-server\tabpy_server\%1
-    SET STARTUP_CMD=%STARTUP_CMD% --config=%1
+    SET STARTUP_CMD=%STARTUP_CMD% --config=%TABPY_ROOT%\tabpy-server\tabpy_server\%1
```
0golovatyi: I am not sure this is what we want. What happens if I run `startup.cmd c:\Users\ogolovatyi\Desktop\myconfig.conf`

johng42: We need to update the docs to specify that this command line parameter will now need to be a relative path (since we are appending/prepending %  to it)
.../docs/server-config.md

sbabayan: I thought the config was restricted to living in tabpy-server/tabpy_server? If not then yeah I don't think this will work

0golovatyi: We can't do that as we adding limitation for users to keep all their config files in one folder and not being able to use them from a shared folder. User should be able to specify either absolute or relative path IMO

tabpy-server/tabpy_server/app/app.py
```diff
@@ -206,7 +206,7 @@ def set_parameter(settings_key,
         # if state.ini does not exist try and create it - remove
         # last dependence on batch/shell script
         set_parameter('state_file_path', ConfigParameters.TABPY_STATE_PATH,
-                      default_val='./', check_env_var=True)
+                      default_val='./tabpy-server/tabpy_server', check_env_var=True)
```
0golovatyi: This breaks unit tests as they are running from the repo root folder.

PR #256, opened 2019-04-10
add test to validate info response with auth off

johng42:
since this actually goes through tornado this should be sufficient for E2E to validate the info response for auth on (already tested) and auth off (this test)

### DIFF
--- a/tabpy-server/server_tests/test_service_info_handler.py
+++ b/tabpy-server/server_tests/test_service_info_handler.py
@@ -129,31 +129,31 @@ class TestServiceInfoHandlerWithoutAuth(AsyncHTTPTestCase):
 
         # create state.ini dir and file
         cls.state_dir = tempfile.mkdtemp(prefix=prefix)
-        with open(os.path.join(cls.state_dir, 'state.ini'), 'w+')
-            as cls.state_file:
+        with open(os.path.join(cls.state_dir, 'state.ini'), 'w+')\
+                as cls.state_file:
             cls.state_file.write('[Service Info]\n'
-                             'Name = TabPy Serve\n'
-                             'Description = \n'
-                             'Creation Time = 0\n'
-                             'Access-Control-Allow-Origin = \n'
-                             'Access-Control-Allow-Headers = \n'
-                             'Access-Control-Allow-Methods = \n'
-                             '\n'
-                             '[Query Objects Service Versions]\n'
-                             '\n'
-                             '[Query Objects Docstrings]\n'
-                             '\n'
-                             '[Meta]\n'
-                             'Revision Number = 1\n')
+                                 'Name = TabPy Serve\n'
+                                 'Description = \n'
+                                 'Creation Time = 0\n'
+                                 'Access-Control-Allow-Origin = \n'
+                                 'Access-Control-Allow-Headers = \n'
+                                 'Access-Control-Allow-Methods = \n'
+                                 '\n'
+                                 '[Query Objects Service Versions]\n'
+                                 '\n'
+                                 '[Query Objects Docstrings]\n'
+                                 '\n'
+                                 '[Meta]\n'
+                                 'Revision Number = 1\n')
         cls.state_file.close()
 
         # create config file
         cls.config_file = tempfile.NamedTemporaryFile(
             prefix=prefix, suffix='.conf', delete=False, mode='w+')
         cls.config_file.write(
-                '[TabPy]\n'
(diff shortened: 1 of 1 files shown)


tabpy-server/server_tests/test_service_info_handler.py
```diff
@@ -120,3 +120,66 @@ def test_given_tabpy_server_with_auth_expect_correct_info_response(self):
...
+                             '\n'
+                             '[Meta]\n'
+                             'Revision Number = 1\n')
+        cls.state_file.close()
+
+        # create config file
+        cls.config_file = tempfile.NamedTemporaryFile(
+            prefix=prefix, suffix='.conf', delete=False)
+        cls.config_file.write(
+            bytes(
```
0golovatyi: You don't have to use `bytes` here - the similar code works without it on line 133.

tabpy-server/server_tests/test_service_info_handler.py
```diff
@@ -120,3 +120,66 @@ def test_given_tabpy_server_with_auth_expect_correct_info_response(self):
...
+        return self.app._create_tornado_web_app()
+
+    def test_tabpy_server_with_no_auth_expect_correct_info_response(self):
+        response = self.fetch('/info')
+        self.assertEqual(response.code, 200)
+        actual_response = json.loads(response.body)
+        expected_response = _create_expected_info_response(
+            self.app.settings, self.app.tabpy_state)
+
+        self.assertDictEqual(actual_response, expected_response)
```
0golovatyi: Probably this one assert is enough here - if anything we check for below is not what we expect this assert will also fail.

johng42: the final assert (~line 183) checks that the returned dictionary is empty.  this is the expected response if there is no pwd file specified, so I intend to leave these asserts unchanged (albeit this is over-validating)

tabpy-server/server_tests/test_service_info_handler.py
johng42: fixed

tabpy-server/server_tests/test_service_info_handler.py
0golovatyi: 👍

tabpy-server/server_tests/test_service_info_handler.py
```diff
@@ -120,3 +120,66 @@ def test_given_tabpy_server_with_auth_expect_correct_info_response(self):
...
+
+
+class TestServiceInfoHandlerWithoutAuth(AsyncHTTPTestCase):
+    @classmethod
+    def setUpClass(cls):
+        prefix = '__TestServiceInfoHandlerWithoutAuth_'
+
+        # create state.ini dir and file
+        cls.state_dir = tempfile.mkdtemp(prefix=prefix)
+        cls.state_file = open(os.path.join(cls.state_dir, 'state.ini'), 'w+')
```
WillAyd: I'd suggest you use a context manager for this, so:

```python
with open(os.path.join(cls.state_dir, 'state.ini'), 'w+') as state_file:
    state_file.write("...")
```

Implicitly closes for you and has better error handling

tabpy-server/server_tests/test_service_info_handler.py
WillAyd: Rather than creating bytes here you could just open the file in `w+` mode to write strings

tabpy-server/server_tests/test_service_info_handler.py
```diff
@@ -120,3 +120,66 @@ def test_given_tabpy_server_with_auth_expect_correct_info_response(self):
...
+        cls.config_file = tempfile.NamedTemporaryFile(
+            prefix=prefix, suffix='.conf', delete=False)
+        cls.config_file.write(
+            bytes(
+                '[TabPy]\n'
+                'TABPY_STATE_PATH = {}'.format(
+                    cls.state_dir),
+            'utf-8'))
+        cls.config_file.close()
+        cls.config_file.close()
```
WillAyd: Not sure if this is intentional but I think calling close twice should raise an error

0golovatyi:
@johng42 build failed

tabpy-server/server_tests/test_service_info_handler.py
johng42: fixed

tabpy-server/server_tests/test_service_info_handler.py
johng42: fixed

PR #257, opened 2019-04-11
add tests for evaluation_plane_handler.py

johng42:


tabpy-server/server_tests/test_evaluation_plane_handler.py
```diff
@@ -67,6 +67,20 @@ def setUpClass(cls):
             '"script":"res=[]\\nfor i in range(len(_arg1)):\\n  '\
             'res.append(_arg1[i] * _arg2[i])\\nreturn res"}'
 
+        cls.script_not_present =\
```
0golovatyi: since these properties are only used once each it may be better just to use value in the according tests directly

johng42: we can refactor if this becomes an issue

Issue #260, opened 2019-04-17
Syntax Error in tabpy-server/tabpy_server/app/util.py


silmor:
In the latest release (0.4) there seems to be a syntax error in tabpy-server/tabpy_server/app/util.py

python tabpy-server/tabpy_server/app/util.py

```
File "tabpy-server/tabpy_server/app/util.py", line 39
    f'The certificate provided expired on {not_after}.',
                                                      ^
SyntaxError: invalid syntax
```

johng42:
https://github.com/tableau/TabPy/pull/261 has the fix for this.  Thanks for the report!

0golovatyi:
@silmor @johng42 How do you repro the issue?

silmor:
@0golovatyi I can reproduce the issue by just cloning the repo and trying to run the start script. Or like described above by directly giving the file tabpy-server/tabpy_server/app/util.py a call.

OS is Linux and macOS, both with the same result.

0golovatyi:
Just tried on Mac, no repro. What Python version are you running?

silmor:
Seems to be related to Python versions < 3.6. I tried a few different versions. From 3.6.0 onwards there no errors.

3.4.0 -> Error
3.5.3 -> Error
3.5.7 -> Error
3.6.0 -> OK
3.6.3 -> OK
3.6.8 -> OK
3.7.3 -> OK

johng42:
We require Python 3.6 and support 3.6.5 or higher.  Your testing seems to bear that out.
The syntax between this logging call and the others does differ, though, so thanks for pointing that out.
I can close this out since your tests worked with supported versions of Python.

WillAyd:
F-strings weren't introduced until Py36 so this is expected and not an issue with TabPy

PR #261, opened 2019-04-17
fix typo

johng42:
issue 260 https://github.com/tableau/TabPy/issues/260

### DIFF
--- a/tabpy-server/tabpy_server/app/util.py
+++ b/tabpy-server/tabpy_server/app/util.py
@@ -37,7 +37,7 @@ def validate_cert(cert_file_path):
     if now > not_after:
         log_and_raise(https_error +
                       'The certificate provided expired on {}.'.format(
-                      not_after),RuntimeError)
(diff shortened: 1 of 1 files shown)


tabpy-server/tabpy_server/app/util.py
```diff
@@ -36,8 +36,8 @@ def validate_cert(cert_file_path):
                           not_before), RuntimeError)
     if now > not_after:
         log_and_raise(https_error +
-                      f'The certificate provided expired on {not_after}.',
-                      RuntimeError)
+                      'The certificate provided expired on {}.'.format(
```
jnegara: Why is this change necessary?

johng42: you get a syntax error when the code is run

johng42: only in the sense that the logging syntax for expired now matches the syntax for not_before

PR #266, opened 2019-04-23
Converted tests to proper idiom

### LINKED ISSUE #215: Refactor test_rest to use self.assertRaises
As mentioned https://github.com/tableau/TabPy/pull/205/files#r264516138 there is currently a lot of tests which intentionally catch exceptions and inspect the type of exception being thrown. This would be more idiomatically expressed by using `self.assertRaises` (this may be an artifact of old 2.X support)

WillAyd:
Closes #215

I converted these tests to a proper `self.assertRaises` idiom. In doing so discovered some of them actually weren't even working as intended (i.e. they weren't raising). I've simply marked those as expectedFailure for now - probably best served in a follow up

### DIFF
--- a/tabpy-tools/tools_tests/test_rest.py
+++ b/tabpy-tools/tools_tests/test_rest.py
@@ -46,23 +46,19 @@ class TestRequestsNetworkWrapper(unittest.TestCase):
             timeout=None,
             auth=None)
 
+    @unittest.expectedFailure
     def test_GET_InvalidData(self):
         url = 'abc'
         data = {'cat'}
-        try:
-            self.assertEqual(self.rnw.GET(url, data), 'json')
-        except Exception:  # TODO: refactor this...
-            e = sys.exc_info()[0]
-            self.assertEquals(e, TypeError)
+        with self.assertRaises(TypeError):
+            self.rnw.GET(url, data)
 
+    @unittest.expectedFailure
     def test_GET_InvalidURL(self):
         url = ''
         data = {'foo': 'bar'}
-        try:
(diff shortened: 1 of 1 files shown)


0golovatyi:
Builds have failed

WillAyd:
Thanks for the heads up. Are the logs exposed outside of your organization? Only thing I saw was a failure of the startup script but no more details beyond that, so not sure if it's actually a failure from the tests or environment configuration

0golovatyi:
@WillAyd No, we don't see anything other than what's shown on https://travis-ci.com/tableau/TabPy/builds/109330713.

Could be Travis glitch, try to start new build maybe. And if it fails again we'll try to repro the failure locally.

WillAyd:
Can you access log file that the startup script creates? From reading back stdout it looks like this fails at `python3 setup.py install` but whatever messages or errors that come from that are redirected by the script to the log file, so the answer lies therein

johng42:
We have a repro and are investigating to see what may have changed to start causing this.
Will, please hold on to this change until we find out what is broken.  Thanks for tipping us off on this!

johng42:
Yesterday urllib3 released a new version, 1.25
we use the requests package, which needs version 1.21.1 of urllib3
so the fix is to require this specific version rather than latest.
We are also adding some extra logging to make this easier to diagnose in the future.
The fix is in dev as I write this and we are porting the required changes to master.
Again, thanks for the report!

jnegara:
We now have a fix in dev and master. Pulling from the latest master should resolve the problem.

0golovatyi:
Looks great, I like it!

PR #274, opened 2019-04-26
Update documentation with more details for auth

0golovatyi:


misc/TabPy.yml
```diff
@@ -261,6 +284,11 @@ definitions:
         type: array
         items:
           type: object
+  UnathorizedError:
```
jnegara: Unauthorized*

PR #279, opened 2019-05-02
call id

0golovatyi:
For request context logging log details once and for all the following log entries just append call id to a message.

### DIFF
--- a/tabpy-server/tabpy_server/app/util.py
+++ b/tabpy-server/tabpy_server/app/util.py
@@ -8,14 +8,6 @@ from OpenSSL import crypto
 logger = logging.getLogger(__name__)
 
 
-def log_and_raise(msg, exception_type):
-    '''
-    Log the message and raise an exception of specified type
-    '''
-    logger.fatal(msg)
-    raise exception_type(msg)
-
-
 def validate_cert(cert_file_path):
     with open(cert_file_path, 'r') as f:
         cert_buf = f.read()
@@ -31,13 +23,15 @@ def validate_cert(cert_file_path):
 
     https_error = 'Error using HTTPS: '
     if now < not_before:
-        log_and_raise(https_error +
-                      'The certificate provided is not valid until {}.'.format(
-                          not_before), RuntimeError)
+        msg = (https_error +
+               f'The certificate provided is not valid until {not_before}.')
+        logger.critical(msg)
+        raise RuntimeError(msg)
     if now > not_after:
-        log_and_raise(https_error +
-                      f'The certificate provided expired on {not_after}.',
-                      RuntimeError)
+        msg = (https_error +
+               f'The certificate provided expired on {not_after}.')
+        logger.critical(msg)
+        raise RuntimeError(msg)
 
 
 def parse_pwd_file(pwd_file_name):
@@ -58,10 +52,10 @@ def parse_pwd_file(pwd_file_name):
     credentials : dict
         Credentials from the file. Empty if succeeded is False.
     '''
-    logger.info('Parsing passwords file {}...'.format(pwd_file_name))
+    logger.info(f'Parsing passwords file {pwd_file_name}...')
 
     if not os.path.isfile(pwd_file_name):
-        logger.fatal('Passwords file {} not found'.format(pwd_file_name))
+        logger.critical(f'Passwords file {pwd_file_name} not found')
         return False, {}
 
     credentials = {}
--- a/tabpy-server/tabpy_server/common/endpoint_file_mgr.py
+++ b/tabpy-server/tabpy_server/common/endpoint_file_mgr.py
@@ -84,7 +84,7 @@ def cleanup_endpoint_files(name, query_path,
     else:
         retain_folders = [os.path.join(local_dir, str(version))
                           for version in retain_versions]
-        logger.log(logging.INFO, f'Retain folder: {retain_folders}')
+        logger.log(logging.INFO, f'Retain folders: {retain_folders}')
 
         for file_or_dir in os.listdir(local_dir):
             candidate_dir = os.path.join(local_dir, file_or_dir)
--- a/tabpy-server/tabpy_server/common/endpoint_file_mgr.py
+++ b/tabpy-server/tabpy_server/common/endpoint_file_mgr.py
@@ -13,25 +13,28 @@ import os
 import shutil
 from re import compile as _compile
 
-logger = logging.getLogger(__name__)
 
 _name_checker = _compile(r'^[a-zA-Z0-9-_\s]+$')
 
 
-def _check_endpoint_name(name):
+def _check_endpoint_name(name, logger=logging.getLogger(__name__)):
     """Checks that the endpoint name is valid by comparing it with an RE and
     checking that it is not reserved."""
     if not isinstance(name, str):
-        log_and_raise("Endpoint name must be a string or unicode", TypeError)
+        msg = 'Endpoint name must be a string or unicode'
+        logger.log(logging.CRITICAL, msg)
+        raise TypeError(msg)
 
     if name == '':
-        log_and_raise("Endpoint name cannot be empty", ValueError)
+        msg = 'Endpoint name cannot be empty'
+        logger.log(logging.CRITICAL, msg)
+        raise ValueError(msg)
 
     if not _name_checker.match(name):
-        log_and_raise(
-            'Endpoint name can only contain: a-z, A-Z, 0-9,'
-            ' underscore, hyphens and spaces.',
-            ValueError)
+        msg = ('Endpoint name can only contain: a-z, A-Z, 0-9,'
+               ' underscore, hyphens and spaces.')
+        logger.log(logging.CRITICAL, msg)
+        raise ValueError(msg)
 
 
 def grab_files(directory):
@@ -51,12 +54,9 @@ def grab_files(directory):
                 yield full_path
 
 
-def get_local_endpoint_file_path(name, version, query_path):
-    _check_endpoint_name(name)
-    return os.path.join(query_path, name, str(version))
-
-
-def cleanup_endpoint_files(name, query_path, retain_versions=None):
+def cleanup_endpoint_files(name, query_path,
+                           logger=logging.getLogger(__name__),
+                           retain_versions=None):
     '''
     Cleanup the disk space a certain endpiont uses.
 
@@ -70,7 +70,7 @@ def cleanup_endpoint_files(name, query_path, retain_versions=None):
         folder for the given version, otherwise, all files for that endpoint
         are removed.
     '''
-    _check_endpoint_name(name)
+    _check_endpoint_name(name, logger=logger)
     local_dir = os.path.join(query_path, name)
 
     # nothing to clean, this is true for state file path where we load
@@ -84,7 +84,7 @@ def cleanup_endpoint_files(name, query_path, retain_versions=None):
     else:
         retain_folders = [os.path.join(local_dir, str(version))
                           for version in retain_versions]
-        logger.info("Retain folder: %s" % retain_folders)
+        logger.log(logging.INFO, f'Retain folder: {retain_folders}')
 
         for file_or_dir in os.listdir(local_dir):
             candidate_dir = os.path.join(local_dir, file_or_dir)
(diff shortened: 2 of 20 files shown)


tabpy-server/tabpy_server/handlers/query_plane_handler.py
```diff
@@ -116,6 +102,22 @@ def _handle_result(self, po_name, data, qry, uid):
...
+            msg = 'Expect input data to be a dictionary'
+            self.logger.log(logging.CRITICAL, msg)
+            raise RuntimeError(msg)
+
+        if "method" in data:
+            return {"data": data.get("data"), "method": data.get("method")}
+        elif "data" in data:
+            return data.get("data")
+        else:
+            msg = ('Expect input data is a dictionary with at least a '
```
johng42: Nit:
 msg = 'Input data must be a dictionary with a key called "data"'

tabpy-server/tabpy_server/handlers/status_handler.py
```diff
@@ -17,8 +14,6 @@ def get(self):
 
         self._add_CORS_header()
 
-        logger.debug(self.append_request_context(
```
johng42: Why remove this?

tabpy-server/tabpy_server/handlers/upload_destination_handler.py
```diff
@@ -18,9 +16,6 @@ def get(self):
             self.fail_with_not_authorized()
             return
 
-        logger.debug(self.append_request_context(
```
johng42: Why remove this?

tabpy-server/tabpy_server/common/endpoint_file_mgr.py
```diff
@@ -84,7 +84,7 @@ def cleanup_endpoint_files(name, query_path, retain_versions=None):
     else:
         retain_folders = [os.path.join(local_dir, str(version))
                           for version in retain_versions]
-        logger.info("Retain folder: %s" % retain_folders)
+        logger.log(logging.INFO, f'Retain folder: {retain_folders}')
```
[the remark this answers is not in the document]
0golovatyi: fixed

tabpy-server/tabpy_server/handlers/query_plane_handler.py
```diff
@@ -116,6 +102,22 @@ def _handle_result(self, po_name, data, qry, uid):
 
             return (None, None)
 
+    def _sanitize_request_data(self, data):
+        if not isinstance(data, dict):
+            msg = 'Expect input data to be a dictionary'
```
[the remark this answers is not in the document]
0golovatyi: fixed

tabpy-server/tabpy_server/handlers/query_plane_handler.py
0golovatyi: fixed

tabpy-server/tabpy_server/handlers/status_handler.py
0golovatyi: base class automatically logs request context which includes method and URL

tabpy-server/tabpy_server/handlers/upload_destination_handler.py
0golovatyi: See comment above

tabpy-server/tabpy_server/app/util.py
```diff
@@ -31,13 +23,15 @@ def validate_cert(cert_file_path):
 
     https_error = 'Error using HTTPS: '
     if now < not_before:
-        log_and_raise(https_error +
-                      'The certificate provided is not valid until {}.'.format(
-                          not_before), RuntimeError)
+        msg = (https_error +
```
WillAyd: If using formatted string literals would be more idiomatic to write this as `f"{https_error}The certificate provided is not valid until {not_before}."`

tabpy-server/tabpy_server/common/endpoint_file_mgr.py
```diff
@@ -13,25 +13,28 @@
 import shutil
 from re import compile as _compile
 
-logger = logging.getLogger(__name__)
 
 _name_checker = _compile(r'^[a-zA-Z0-9-_\s]+$')
 
 
-def _check_endpoint_name(name):
+def _check_endpoint_name(name, logger=logging.getLogger(__name__)):
```
WillAyd: Not sure the intention of making this a keyword argument but probably better to just initialize the logger in the global namespace of your module rather than defaulting it locally to this function

tabpy-server/tabpy_server/common/endpoint_file_mgr.py
```diff
@@ -13,25 +13,28 @@
...
     """Checks that the endpoint name is valid by comparing it with an RE and
     checking that it is not reserved."""
     if not isinstance(name, str):
-        log_and_raise("Endpoint name must be a string or unicode", TypeError)
+        msg = 'Endpoint name must be a string or unicode'
```
WillAyd: With Py2 being dropped shouldn't reference unicode anymore

tabpy-server/tabpy_server/common/endpoint_file_mgr.py
```diff
@@ -84,7 +84,7 @@ def cleanup_endpoint_files(name, query_path, retain_versions=None):
     else:
         retain_folders = [os.path.join(local_dir, str(version))
                           for version in retain_versions]
-        logger.info("Retain folder: %s" % retain_folders)
+        logger.log(logging.INFO, f'Retain folders: {retain_folders}')
```
WillAyd: Just call `logger.info`

tabpy-server/tabpy_server/handlers/base_handler.py
```diff
@@ -1,30 +1,143 @@
...
+import uuid
+
 
-logger = logging.getLogger(__name__)
 STAGING_THREAD = concurrent.futures.ThreadPoolExecutor(max_workers=3)
 
 
-class BaseHandler(tornado.web.RequestHandler):
-    KEYS_TO_SANITIZE = ("api key", "api_key", "admin key", "admin_key")
+class ContextLoggerWrapper(object):
```
WillAyd: Declaring classes like this is also an artifact from new / old style classes in Python2. Since you don't support that any more would give your codebase a more modern look if you just did `class ContextLoggerWrapper` (no functionality will change)

tabpy-server/tabpy_server/handlers/base_handler.py
```diff
@@ -1,30 +1,143 @@
...
+    '''
+    This class appends request context to logged messages.
+    '''
+    @staticmethod
+    def _generate_call_id():
+        return str(uuid.uuid4())
+
+    def __init__(self, request: tornado.httputil.HTTPServerRequest):
```
WillAyd: Love the types! Any plans to add MyPy to CI soon?

tabpy-server/tabpy_server/handlers/base_handler.py
```diff
@@ -1,30 +1,143 @@
...
+    def set_request(self, request: tornado.httputil.HTTPServerRequest):
+        '''
+        Set HTTP(S) request for logger. Headers will be used to
+        append request data as client information, Tableau user name, etc.
+        '''
+        self.remote_ip = request.remote_ip
+        self.method = request.method
+        self.url = request.full_url()
+
+        if 'TabPy-Client' in request.headers:
```
WillAyd: More idiomatic to write this as:

`self.client = request.headers.get('TabPy-Client')`

tabpy-server/tabpy_server/handlers/base_handler.py
```diff
@@ -1,30 +1,143 @@
...
+
+        if 'TabPy-User' in request.headers:
+            self.tableau_username = request.headers['TabPy-User']
+        else:
+            self.tableau_username = None
+
+    def set_tabpy_username(self, tabpy_username: str):
+        self.tabpy_username = tabpy_username
+
+    def enable_context_logging(self, enable: bool):
```
WillAyd: Just a suggestion but whenever mutating objects without returning anything it's clearer IMO to name the method `set_...` 

With that said with how simple this is you probably don't need a method and can just assign to `log_request_context` directly

tabpy-server/tabpy_server/handlers/base_handler.py
```diff
@@ -1,30 +1,143 @@
...
+            Same as kwargs in Logger.debug().
+        '''
+        extended_msg = msg
+        if self.log_request_context:
+            if not self.request_context_logged:
+                self._log_context_info()
+
+            extended_msg += f', <<call ID: {self.call_id}>>'
+
+        logging.getLogger(__name__).log(level, extended_msg)
```
WillAyd: Again would suggest assigning `logger = logging.getLogger(__name__)` in the global namespace; it's done most other places in the code base so would help with consistency

tabpy-server/tabpy_server/handlers/base_handler.py
```diff
@@ -96,75 +205,85 @@ def _get_auth_method(self, api_version) -> (bool, str):
         is not needed.
         '''
         if api_version not in self.settings[SettingsParameters.ApiVersions]:
-            logger.critical(f'Unknown API version "{api_version}"')
+            self.logger.log(logging.CRITICAL,
```
WillAyd: `self.logger.critical`

tabpy-server/tabpy_server/handlers/base_handler.py
```diff
@@ -96,75 +205,85 @@ def _get_auth_method(self, api_version) -> (bool, str):
...
+                            f'Unknown API version "{api_version}"')
             return False, ''
 
         version_settings =\
             self.settings[SettingsParameters.ApiVersions][api_version]
         if 'features' not in version_settings:
-            logger.info(f'No features configured for API "{api_version}"')
+            self.logger.log(logging.INFO,
```
WillAyd: `self.logger.info`

tabpy-server/tabpy_server/handlers/base_handler.py
```diff
@@ -96,75 +205,85 @@ def _get_auth_method(self, api_version) -> (bool, str):
...
+                            f'No features configured for API "{api_version}"')
             return True, ''
 
         features = version_settings['features']
         if 'authentication' not in features or\
                 not features['authentication']['required']:
-            logger.info(
-                f'Authentication is not a required feature for API '
-                '"{api_version}"')
+            self.logger.log(
```
WillAyd: and so on...

tabpy-server/tabpy_server/app/util.py
0golovatyi: Yes, eventually we want to use formatted strings everywhere as they are easy to read and safer. In this change, I missed some places where the strings can be updated... Some next change maybe :)

tabpy-server/tabpy_server/common/endpoint_file_mgr.py
0golovatyi: Depending on the context the function can be called from HTTP request handler and then `logger` will be a wrapper which appends call ID to a logged message. Or it can be called from an app (e.g. when deploying a model) and then it uses the standard logger.

tabpy-server/tabpy_server/common/endpoint_file_mgr.py
0golovatyi: Yes, we really need to clean up Unicode references everywhere. Not with this change though.

tabpy-server/tabpy_server/common/endpoint_file_mgr.py
0golovatyi: If `logger` is a call context logger wrapper it won't have `info` method

tabpy-server/tabpy_server/handlers/base_handler.py
0golovatyi: Yeah, I don't like exposing the internals of a type to client code, this is why I have the method. As for the name of if `set_context_logging_enabled` seems too long for me.

tabpy-server/tabpy_server/handlers/base_handler.py
0golovatyi: logger wrapper doesn't have method `critical`

tabpy-server/tabpy_server/handlers/base_handler.py
0golovatyi: This is exactly what I want to avoid - global logger :)

PR #294, opened 2019-05-14
Add a warning when python version is below 3.6.5

rxing-tableau:
A fix for bug
[TabPy] we don't warn user if using python 2.7x that tabpy will fail to start

### DIFF
--- a/startup.cmd
+++ b/startup.cmd
@@ -11,7 +11,7 @@ SET SAVE_PYTHONPATH=%PYTHONPATH%
 ECHO Checking for presence of Python in the system path variable.
 python --version
 IF %ERRORLEVEL% NEQ 0 (
-    ECHO     Cannot find Python.exe.  Check that Python is installed and is in the system PATH environment variable.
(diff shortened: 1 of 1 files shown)


startup.cmd
```diff
@@ -14,6 +14,14 @@ IF %ERRORLEVEL% NEQ 0 (
     ECHO     Cannot find Python.exe.  Check that Python is installed and is in the system PATH environment variable.
     SET RET=1
     GOTO:END
+) ELSE (
+    FOR /F "TOKENS=2" %%a IN ('python --version') DO (
```
0golovatyi: How will this work in Anaconda where `python --version` returns something like `Python 3.6.8 :: Anaconda, Inc.`?

rxing-tableau: Sorry I closed this request, because I want to combine the change with "startup.sh". I will send out a new request later.
But yes, let's clarify your question first.
Since I am retrieving the second token (tokens=2), it will treat the version as 3.6.8 which is higher than 3.6.5. Is that what we need?

0golovatyi: Yes, we want to check for 3.6.5 at least.

PR #296, opened 2019-05-16
Print a warning/error if python version is not as required

rxing-tableau:
Change: 
If python version is below 3.6, print an error to console, and block the start up, since it won't success anyway.
If python version is below 3.6.5, print a warning to console, without blocking the start up.

Related Task: 
https://tfs.tsi.lan/tfs/DefaultCollection/Tableau-Dev/Advanced%20Analytics%20-%20(Augmented%20Analytics%20Area)/_workitems?id=913569&triage=true&_a=edit

### DIFF
--- a/startup.cmd
+++ b/startup.cmd
@@ -24,7 +24,7 @@ IF %ERRORLEVEL% NEQ 0 (
             SET RET=1
             GOTO:END
         ) ELSE IF %%a LSS %DESIRED_PY_VER% (
-            ECHO Important Warning : Python %%a% is not supported. Please upgrade Python to 3.6.5 or higher.
+            ECHO Warning : Python %%a% is not supported. Please upgrade Python to 3.6.5 or higher.
             SET RET=1
         )
     )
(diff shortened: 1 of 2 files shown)


startup.cmd
```diff
@@ -6,14 +6,28 @@ REM Set environment variables.
...
     SET RET=1
     GOTO:END
+) ELSE (
+    FOR /F "TOKENS=2" %%a IN ('python --version 2^>^&1') DO (
+        IF %%a LSS %MIN_PY_VER% (
+            ECHO %PYTHON_ERROR%
+            SET RET=1
+            GOTO:END
+        ) ELSE IF %%a LSS %DESIRED_PY_VER% (
+            ECHO Important Warning : Python %%a% is not supported. Please upgrade Python to 3.6.5 or higher.
```
johng42: I suggest removing the word "Important" to better match our logging text.

rxing-tableau: I remove "Important"

Issue #118, opened 2018-07-13
hard coded 30 sec readtimeout


sajozsattila:
On TabPy server it drop readtimeout after 30 sec.  This is hardcoded on the 565 line.  I guess it could be good if it can be set from the config file.

Patch for this:
```
--- tabpy-server/tabpy_server/tabpy.py	2018-07-11 12:33:41.306496157 +0000
+++ /opt/conda/envs/Tableau-Python-Server/lib/python3.6/site-packages/tabpy_server/tabpy.py	2018-07-11 13:58:54.785388360 +0000
@@ -109,6 +109,7 @@
         # set content type to application/json
         self.set_header("Content-Type", "application/json")
         self.port = self.settings['port']
+        self.timeout = self.settings['timeout']
         self.py_handler = self.settings['py_handler']
 
     def error_out(self, code, log_message, info=None):
@@ -538,7 +539,7 @@
 
     @gen.coroutine
     def call_subprocess(self, function_to_evaluate, arguments):
-        restricted_tabpy = RestrictedTabPy(self.port)
+        restricted_tabpy = RestrictedTabPy(self.port, self.timeout)
         # Exec does not run the function, so it does not block.
         if sys.version_info > (3, 0):
             exec(function_to_evaluate, globals())
@@ -555,8 +556,9 @@
 
 
 class RestrictedTabPy:
-    def __init__(self, port):
+    def __init__(self, port, timeout):
         self.port = port
+        self.timeout = timeout
 
     def query(self, name, *args, **kwargs):
         url = 'http://localhost:%d/query/%s' % (self.port, name)
@@ -564,7 +566,7 @@
         data = simplejson.dumps(internal_data)
         headers = {'content-type': 'application/json'}
         response = requests.post(url=url, data=data, headers=headers,
-                                 timeout=30)
+                                 timeout=self.timeout)
 
         return response.json()
 
@@ -802,6 +804,12 @@
         settings['upload_dir'] = os.getenv('TABPY_QUERY_OBJECT_PATH',
                                            '/tmp/query_objects')
 
+    try:
+        settings['timeout'] = int(config.TABPY_TIMEOUT)
+    except AttributeError:
+        settings['timeout'] = int(os.getenv('TABPY_TIMEOUT', 30))
+
+
     if not os.path.exists(settings['upload_dir']):
         os.makedirs(settings['upload_dir'])
```

jnegara:
Thanks for the feedback! We implemented your suggestion in PR #298.

PR #298, opened 2019-05-21
Added parameter for configuring timeout for /evaluate.

jnegara:
Added support for parameter `TABPY_EVALUATE_TIMEOUT` in the server config.

When `/evaluate` is called with some script, if that script takes longer than the number of seconds specified by this parameter the script is terminated and the client received a `TimeoutError`.

If `TABPY_EVALUATE_TIMEOUT` is set to a non-float value, or is not present then the timeout length defaults to 30s.

This should close issue #118.


docs/server-config.md
```diff
@@ -192,3 +193,12 @@ arg1, _arg2):
 No passwords are logged.
 
 NOTE the request context details are logged with INFO level.
+
+## Custom Script Timeout
+
+By default, all custom scripts execute through `POST /evaluate` may run for up
```
sbabayan: executed*

docs/server-config.md
```diff
@@ -192,3 +193,12 @@ arg1, _arg2):
...
+to 30.0 s before being terminated. To configure this timeout, uncomment
+`TABPY_EVALUATE_TIMEOUT = 30` in the default config under
+`tabpy-server/tabpy_server/common/default.conf` and replace `30` with the float
+value of your choice representing the timeout time in seconds, or add an such
```
sbabayan: or add such*

tests/integration/test_custom_evaluate_timeout.py
```diff
@@ -0,0 +1,57 @@
...
+        self._run_test(payload)
+
+    def _run_test(self, payload):
+        conn = self._get_connection()
+        conn.request('POST', '/evaluate', payload, self.headers)
+        res = conn.getresponse()
+        res_bytes = res.read().decode('utf-8')
+
+        self.assertEqual(500, res.status)
+        print('res_bytes: ' + res_bytes)
```
sbabayan: Why does the integration test need to print?

tabpy-server/tabpy_server/handlers/evaluation_plane_handler.py
```diff
@@ -5,22 +5,24 @@
...
+        self.timeout = timeout
 
     def query(self, name, *args, **kwargs):
         url = f'http://localhost:{self.port}/query/{name}'
         self.logger.log(logging.DEBUG, f'Querying {url}...')
         internal_data = {'data': args or kwargs}
         data = json.dumps(internal_data)
         headers = {'content-type': 'application/json'}
+
+        print(f'timeout = {self.timeout}')
```
0golovatyi: remove this - debugging artifact?

tests/integration/test_custom_evaluate_timeout.py
jnegara: it doesn't, good catch

tabpy-server/tabpy_server/handlers/evaluation_plane_handler.py
jnegara: yup

PR #299, opened 2019-05-23
TFSID 924825: When attempting to call a deployed model from TabPy get "EvaluationPlaneHandler.py has no attribute 'log'"

sbabayan:
fixed bug where we cannot access model from tableau, extended integration test to cover this scenario


0golovatyi:
We need to delete releases 0.4.1 and 0.5 as they have the bug in them - please update VERSION and changelog.

sbabayan:
Closed this pr until we can fix configurable timeout bug, will merge just the models fix to master for now (in a new pr)

PR #301, opened 2019-05-24
TFSID 924825: When attempting to call a deployed model from TabPy get "EvaluationPlaneHandler.py has no attribute 'log'"

sbabayan:
fixed bug where we cannot access model from tableau, extended integration test to cover this scenario


0golovatyi:
Also seems like new unit test is needed to increase code coverage

models/scripts/SentimentAnalysis.py
```diff
@@ -1,6 +1,9 @@
 from tabpy_tools.client import Client
 from textblob import TextBlob
-from nltk.sentiment import SentimentIntensityAnalyzer
+import nltk
+nltk.download('vader_lexicon')
```
sbabayan: Why are these downloads necessary? Does pip installing nltk not install everything that's needed?

tests/integration/integ_test_base.py
```diff
@@ -150,7 +150,7 @@ def _get_config_file_name(self) -> str:
         config_file = open(os.path.join(self.tmp_dir, 'test.conf'), 'w+')
         config_file.write(
             '[TabPy]\n'
-            f'TABPY_QUERY_OBJECT_PATH = {self.tmp_dir}/query_objects\n'
+            f'TABPY_QUERY_OBJECT_PATH = ./query_objects\n'
```
sbabayan: If we create this directory in the root won't it not get cleaned up after the test run?

models/scripts/SentimentAnalysis.py
0golovatyi: didn't work for me without that

tests/integration/integ_test_base.py
0golovatyi: the folder is created in the temp folder which is current folder when the test runs

models/scripts/SentimentAnalysis.py
nmannheimer: Yes the download is necessary. nltk does not install the vader lexicon by defaul, which is required for sentiment analysis in nltk. It's essentially a text file that contains the known sentiment scores of various words.

tests/integration/integ_test_base.py
sbabayan: I see, so we don't actually need to specify {self.tmp_dir} for TABPY_STATE_PATH either?

0golovatyi: Probably not. We change the folder either when starting TabPy with startup script or when pointing to config path.

tests/integration/test_deploy_and_evaluate_model.py
```diff
@@ -0,0 +1,22 @@
+import integ_test_base
+import subprocess
+from pathlib import Path
+
+
+class TestDeployModelSSLOffAuthOff(integ_test_base.IntegTestBase):
```
0golovatyi: This class name is already used in another test file.

sbabayan: thanks for catching that i forgot to update it

0golovatyi:
Still need to increase unit tests coverage.

PR #306, opened 2019-05-29
fixing regex issue where . were not allowed in endpoint path (#305)

sbabayan:
merging fix into dev

### DIFF
--- a/tabpy-server/tabpy_server/handlers/management_handler.py
+++ b/tabpy-server/tabpy_server/handlers/management_handler.py
@@ -94,7 +94,7 @@ class ManagementHandler(MainHandler):
                 self.settings[SettingsParameters.StateFilePath], name, version)
             self.logger.log(logging.DEBUG,
                             f'Checking source path {src_path}...')
-            _path_checker = _compile(r'^[\\\:a-zA-Z0-9-_~\s/]+$')
+            _path_checker = _compile(r'^[\\\:a-zA-Z0-9-_~\s/\.]+$')
             # copy from staging
             if src_path:
                 if not isinstance(request_data['src_path'], str):


tabpy-server/tabpy_server/handlers/management_handler.py
```diff
@@ -94,7 +94,7 @@ def _add_or_update_endpoint(self, action, name, version, request_data):
                 self.settings[SettingsParameters.StateFilePath], name, version)
             self.logger.log(logging.DEBUG,
                             f'Checking source path {src_path}...')
-            _path_checker = _compile(r'^[\\\:a-zA-Z0-9-_~\s/]+$')
+            _path_checker = _compile(r'^[\\\:a-zA-Z0-9-_~\s/\.]+$')
```
WillAyd: Not exactly sure of the issue you are trying to solve but I'm somewhat surprised this is the issue; special characters don't hold meaning within a character set.

I'm a little wary generally of what's going on in here. You might want to move the `-` after the `9` to the end of the regex; otherwise that can easily be misinterpreted as a range instead of a literal dash. Also since you are using a raw string you *should* only need `\\` at the start of the character set to match a literal backslash instead of `\\\`, which may be trying to escape the `:` immediately following it

jnegara: Regarding the issue this PR is trying to solve, a recent change to this regex pattern introduced a bug in which models could not be deployed if the state file path contained a `.` character in it.

WillAyd:
Gotcha. Wouldn’t this work without the backslash though?

jnegara:
Haha maybe. I'm not really good at regex, but it seems to work when I test it. @0golovatyi thoughts?

tabpy-server/tabpy_server/handlers/management_handler.py
0golovatyi: @WillAyd I agree validating query object path with regex doesn't make much sense as an OS already rejects invalid path. We may remove the whole validation piece in the future but at the moment not looking to do any refactoring with the change.

As for the regex itself: first `\\` needed to escape `\`, then `\:` needed to escape `:` as it has a special meaning inside `[]`. And `\.` is needed as we are allowing `.` character, not just any character.

jnegara: I opened an issue to address the regex path validation. #308

PR #316, opened 2019-06-15
Use Native Coroutines instead of gen decorator

WillAyd:
Unpins older tornado version with updates as suggested by Tornado:

https://www.tornadoweb.org/en/stable/guide/coroutines.html#native-vs-decorated-coroutines

### DIFF
--- a/tabpy-server/setup.py
+++ b/tabpy-server/setup.py
@@ -47,7 +47,6 @@ setup(
         'singledispatch',
         'six',
         'tornado',
-        'Tornado-JSON',
         'urllib3<1.25,>=1.21.1'
     ]
 )
--- a/tabpy-server/setup.py
+++ b/tabpy-server/setup.py
@@ -46,7 +46,7 @@ setup(
         'requests',
         'singledispatch',
         'six',
-        'tornado==5.1.1',
+        'tornado',
         'Tornado-JSON',
         'urllib3<1.25,>=1.21.1'
     ]
--- a/tabpy-server/tabpy_server/handlers/evaluation_plane_handler.py
+++ b/tabpy-server/tabpy_server/handlers/evaluation_plane_handler.py
@@ -34,7 +34,7 @@ class EvaluationPlaneHandler(BaseHandler):
         super(EvaluationPlaneHandler, self).initialize(app)
         self.executor = executor
 
-    async def post(self):
+    def post(self):
         if self.should_fail_with_not_authorized():
             self.fail_with_not_authorized()
             return
@@ -77,8 +77,8 @@ class EvaluationPlaneHandler(BaseHandler):
                 logging.INFO,
                 f'function to evaluate={function_to_evaluate}')
 
-            result = await self.call_subprocess(function_to_evaluate,
-                                                arguments)
+            result = yield from self.call_subprocess(function_to_evaluate,
+                                                     arguments)
             if result is None:
                 self.error_out(400, 'Error running script. No return value')
             else:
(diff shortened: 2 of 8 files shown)


tabpy-server/tabpy_server/psws/callbacks.py
```diff
@@ -36,7 +35,6 @@ def wait_for_endpoint_loaded(python_service, object_uri):
         sleep(0.1)
 
 
-@gen.coroutine
```
WillAyd: Items in this module didn't appear to be doing anything asynchronous so simply removed the decorator and did not add async in contrast to other methods

WillAyd:
Hmm so error message seems to be a result of mixing but multithreaded and asynchronous operations. Is there any particular reason to have both in the application or would a port to entirely asynchronous be considered?

tabpy-server/setup.py
```diff
@@ -46,8 +46,7 @@
         'requests',
         'singledispatch',
         'six',
-        'tornado==5.1.1',
-        'Tornado-JSON',
+        'tornado',
```
rxing-tableau: We need to guarantee all the tests pass before upgrade tornado to the latest version.
Can't just install the latest one.

tabpy-server/tabpy_server/handlers/evaluation_plane_handler.py
```diff
@@ -112,5 +109,4 @@ def call_subprocess(self, function_to_evaluate, arguments):
         else:
             future = self.executor.submit(_user_script, restricted_tabpy,
                                           **arguments)
-        ret = yield future
-        raise gen.Return(ret)
+        return asyncio.wrap_future(future)
```
rxing-tableau: Why not ?
async def call_subprocess(self, function_to_evaluate, arguments):
...
return await future

tabpy-server/setup.py
WillAyd: What version do you want to pin it to?

WillAyd:
Thanks for the responses! I think one of them might have gotten lost in GitHub somewhere but in regards to the `call_subprocess` method I don't think it can be awaited because that actually returns a Future from a multi-threaded operation rather than an a Task.

Is there any particular reason to use multi-threaded operations or could we stick all asynchronous? Seems like the latter would make for a more unified code base with easier maintenance so happy to invest time in making that happen just not sure if there is a historical reason both paradigms were intertwined.

rxing-tableau:
Thank you for all the contributions Will.
As I know, there is no historical blocker from moving to unified asynchronous paradigm. It just needs the refactor.
For the Tornado version, we can only support 5.1.1 now, because there is known issue when we move to higher version. It probably related with the gen.coroutines decorator.

WillAyd:
> For the Tornado version, we can only support 5.1.1 now, because there is known issue when we move to higher version. It probably related with the gen.coroutines decorator.

Oh yea! So I should have been more explicit but that is the point of this. If we remove these decorators and place with the native async syntax from Python (as suggested by Tornado anyway) we wouldn't need to pin to 5.1. A lot of those features were intentionally removed in Tornado 6

> As I know, there is no historical blocker from moving to unified asynchronous paradigm. It just needs the refactor.

Good to know. I'll take a look at removing that altogether over the next few days then - just doing async should be easier to maintain and would be less buggy

0golovatyi:
@WillAyd Do you still have plans to work on this change? If not I am going to close the PR.

WillAyd:
Feel free to close for now - can always pick up later

PR #328, opened 2019-07-30
Upgrade to Tornado 6

0golovatyi:
With this change TabPy is not locked to Tornado 5.1.1 anymore and can use the latest version.


CHANGELOG
```diff
@@ -3,6 +3,14 @@
...
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## v0.7
+
+### Improvements
+
+- Added t-test model
+- Fixed models call with /evaluate for HTTPS
+- Migrated to Tornado 6
+
```
sbabayan: Didn't we also bring over jessica's timeout work?

docs/FAQ.md
```diff
@@ -11,4 +11,5 @@
 
 ### AttributeError: module 'tornado.web' has no attribute 'asynchronous'
 
-TabPy uses Tornado 5.1.1. To it to your Python environment run `pip install tornado==5.1.1` and then try to start TabPy again.
+TabPy uses Tornado 5.1.1. To it to your Python environment run
```
sbabayan: If we are no longer bound to 5.1.1 should we remove this from the documentation?

0golovatyi: Yes, we should. But I'll keep it for now for 2 reasons:
1. People may still have the issue using an older version of TabPy, and
2. The file itself will be empty before we add any other Q/A there :)

CHANGELOG
0golovatyi: Updating changelog...

PR #335, opened 2019-08-15
Correct Path reference to custom config file

### LINKED ISSUE #334: Incorrect Documentation in Regards to Config File Location
**Environment information:**
 - OS: Microsoft Server 2016
 - Python version: 3.7.1
 - TabPy release: 0.6.1

**Describe the issue**
Relative path for config file in [documentation](https://github.com/tableau/TabPy/blob/master/docs/server-startup.md#command-line-arguments-for-windows) appears incorrect.  

**To Reproduce**
Per documentation I placed a custom config file at: 
`%TABPY_ROOT%\tabpy-server\tabpy_server\myconfig.conf`

When starting tabpy with `startup.cmd myconfig.conf` i get an error that it cannot find the custom config file and is reverting to default.

**Additional context**
As troubleshooting I moved the location of the myconfig.conf file and moved it to the `%TABPY_ROOT%` folder.  When I run `startup.cmd myconfig.conf` with the custom file in `%TABPY_ROOT%` it finds the file and starts without problems, but this seems contrary to what is the current docs.

jtweeder:
Correct documentation to show the correct relative path for providing a custom configuration file at the command line start up.

Fixes #334

### DIFF
--- a/docs/server-startup.md
+++ b/docs/server-startup.md
@@ -57,10 +57,10 @@ startup.cmd myconfig.conf
 ```
 
 Replace `myconfig.conf` with the path to your config file relative to
-`%TABPY_ROOT%\tabpy-server\tabpy_server\`.
+`%TABPY_ROOT%\`.
 
 For example, in this case your config file would be located at
-`%TABPY_ROOT%\tabpy-server\tabpy_server\myconfig.conf`.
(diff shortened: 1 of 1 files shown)


jnegara:
Good catch, thanks for your contribution!

PR #346, opened 2019-10-01
Enable config file to parse environment variables.

jnegara:
Environment variables can now be used in the config file such that `%(ENV_VAR)s` will be parsed as the value of the environment variable `ENV_VAR`.

### DIFF
--- a/tabpy/tabpy_server/app/app.py
+++ b/tabpy/tabpy_server/app/app.py
@@ -183,7 +183,7 @@ class TabPyApp:
                 logger.debug(
                     f'Parameter {settings_key} set to '
                     f'"{self.settings[settings_key]}" '
-                    'from config file')
+                    'from config file or environment variable')
 
             if not key_is_set and default_val is not None:
                 self.settings[settings_key] = default_val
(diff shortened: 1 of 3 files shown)


tabpy/tabpy_server/app/app.py
```diff
@@ -160,7 +160,7 @@ def _parse_config(self, config_file):
         self.python_service = None
         self.credentials = {}
 
-        parser = configparser.ConfigParser()
+        parser = configparser.ConfigParser(os.environ)
```
0golovatyi: In `set_parameter` we specifically check for environment variables with the same names and config file parameters. Do we remove that check now?

jnegara: It looks to me like we only look for environment variables if the parameter is not set in the config.

https://github.com/tableau/TabPy/blob/dev/tabpy/tabpy_server/app/app.py#L179

jnegara: Misunderstood your comment - as a side effect of my change we no longer need to check for environment variables.

Removed the additional check.

jnegara:
For the record, I am seeing this same test failure in the dev branch.
https://travis-ci.com/tableau/TabPy/builds/124143059

PR #375, opened 2019-12-15
Change regex, add remove method, and edit qeury_timeout

phantomcosmonaut:
Unit tests all passed. I modified your regex to be more compact and made some updates to docstrings. It also looked like the query_timeout attribute was not being set correctly because it was initially set to self but in the property wrapper it was being set to self._service and the latter seems correct. I also made the query_timeout parameter more robust by filtering for argument type before setting. In a previous issue, someone brought up that there was no remove endpoint function so I added one. It works: '[INFO] (web.py:web:2106): 204 DELETE /endpoints/Foo (127.0.0.1) 9.97ms', but it appears to cause a server-side error: '[ERROR] (base_handler.py:base_handler:120): Error submitting update model request: error=AttributeError : 'ContextLoggerWrapper' object has no attribute 'info''. However, client.get_status() is resolved when the server is restarted.


0golovatyi:
@phantomcosmonaut  The change looks good. I suggest to merge the change to the master branch instead - dev branch is where development by Tableau happens. It may take some time before dev is merged to master and a new release is created.

phantomcosmonaut:
I originally cloned the dev branch thinking I'd push changes to that :/ so if I submit a new pull request on your master branch it will include other modifications on that file I did not make.  Do you want me to go ahead with that or clone the master branch and make my changes?

0golovatyi:
It is fine, we can bring the change to dev branch. The only thing - I can't tell when the fix will make it to a release and package.

tabpy/tabpy_tools/client.py
```diff
@@ -22,14 +22,13 @@ def _check_endpoint_type(name):
 
 def _check_hostname(name):
     _check_endpoint_type(name)
-    hostname_checker = compile(
-        r'^http(s)?://[a-zA-Z0-9-_\.]+(/)?(:[0-9]+)?(/)?$')
+    hostname_checker = compile(r'^http(s)?://[\w.-]+(/)?(:\d+)?(/)?$')
```
WillAyd: Just out of curiosity is there a reason to use this regex instead of `urllib.parse`?

https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse

phantomcosmonaut: That's a good idea. It could add more functionality if needed but it's probably just regex under the hood right?

WillAyd: I don't know the details but can find here:

https://github.com/python/cpython/blob/f501db2b93a9d3d840b6fb38d6bdda8bcc400d4a/Lib/urllib/parse.py#L366

unless needed to roll separately here I just think a better idea to use built-ins. One gap in this implementation is for ipv6 addresses, but ultimately up to the Tableau team

0golovatyi: I think we should prefer a dedicated URL parser/validator even if it is implemented with exactly the same regex.

PR #377, opened 2019-12-16
Convert NaNs to Nulls during json serialization

tfoldi:
`NaN` values in arrays cause serialization issues in Desktop/Server. All `NaN` should be converted to `null` when returning from rest api


tabpy/tabpy_server/handlers/evaluation_plane_handler.py
```diff
@@ -94,7 +95,7 @@ def post(self):
             if result is None:
                 self.error_out(400, 'Error running script. No return value')
             else:
-                self.write(json.dumps(result))
+                self.write(simplejson.dumps(result, ignore_nan=True))
```
0golovatyi: Trailing whitespaces.

tfoldi:
I'm still not happy with the change, as I guess we have to change all json dumps to `simplejson` from `json`. What do you think? 

Maybe a more generic solution could be to do:

```
try:
    import simplejson as json
except ImportError:
    import json
```

Plus, do you need any test case to reproduce the `NaN` issue or you guys just believe me (which is perfectly fine)

tabpy/tabpy_server/handlers/evaluation_plane_handler.py
```diff
@@ -94,7 +95,7 @@ def post(self):
...
```
0golovatyi: There is `allow_nan` parameter in `json.dumps`.

tfoldi: that's cool, but then what? will it replace NaNs to nulls?

tfoldi: > If allow_nan is true (the default), then NaN, Infinity, and -Infinity will be encoded as such. This behavior is not JSON specification compliant, but is consistent with most JavaScript based encoders and decoders. Otherwise, it will be a ValueError to encode such floats.

as I understood, this is not the behavior we need. In fact, `ignore_nan` in `simplejson` does exactly what we want:

> If ignore_nan is true (default: False), then out of range float values (nan, inf, -inf) will be serialized as null in compliance with the ECMA-262 specification. If true, this will override allow_nan.

0golovatyi: Makes sense. You were considering adding a unit or an integration test for the scenario?

PR #379, opened 2019-12-17
Ran black on project

### LINKED ISSUE #364: Code Style: black
Quite a few open source projects now use black from the PSF for code formatting. In essence, it does all code formatting for you so you don't have to think about it

https://github.com/psf/black

I think would be good for TabPy to clean up code. From experience you can just run this wide open on your code base, commit the result and move on from there, probably setting up in CI as well

WillAyd:
closes #364

After this would probably want to set up a CI rule to enforce. GitHub actions has a few if you wanted to use any of those:

https://github.com/marketplace?utf8=✓&type=actions&query=black

Otherwise could add to existing CI


WillAyd:
Hmm yea one other conflict - flake8 defaults to 79 characters per line whereas black is 88. I typically see people update the flake8 config to allow 88 but can go either way

0golovatyi:
@WillAyd we should have flake8/pep8 config file in the repo where the restrictions can be configured. 88 or even more characters per line seems more reasonable than 78 IMO.

tabpy/tabpy_server/app/util.py
```diff
@@ -51,42 +47,41 @@ def parse_pwd_file(pwd_file_name):
...
-                    f'Incorrect entry "{row}" in password file')
+                logger.error(f'Incorrect entry "{row}" in password file')
                 return False, {}
 
             login = row[0].lower()
             if login in credentials:
                 logger.error(
-                    f'Multiple entries for username {login} '
-                    'in password file')
+                    f"Multiple entries for username {login} " "in password file"
```
0golovatyi: These 2 strings should be concatenated.

tabpy/tabpy_server/handlers/__init__.py
```diff
@@ -4,10 +4,10 @@
...
 from tabpy.tabpy_server.handlers.endpoints_handler import EndpointsHandler
-from tabpy.tabpy_server.handlers.evaluation_plane_handler\
-    import EvaluationPlaneHandler
+from tabpy.tabpy_server.handlers.evaluation_plane_handler import EvaluationPlaneHandler
 from tabpy.tabpy_server.handlers.query_plane_handler import QueryPlaneHandler
 from tabpy.tabpy_server.handlers.service_info_handler import ServiceInfoHandler
 from tabpy.tabpy_server.handlers.status_handler import StatusHandler
-from tabpy.tabpy_server.handlers.upload_destination_handler\
-    import UploadDestinationHandler
+from tabpy.tabpy_server.handlers.upload_destination_handler import (
```
0golovatyi: Seems like () and ,  can be removed

tabpy/tabpy_server/handlers/evaluation_plane_handler.py
```diff
@@ -45,88 +47,90 @@ def post(self):
...
+            arguments_str = ""
+            if "data" in body:
+                arguments = body["data"]
 
             if arguments is not None:
                 if not isinstance(arguments, dict):
-                    self.error_out(400, 'Script parameters need to be '
-                                        'provided as a dictionary.')
+                    self.error_out(
+                        400, "Script parameters need to be " "provided as a dictionary."
```
0golovatyi: Strings should be concatenated.

tabpy/tabpy_server/app/util.py
WillAyd: Nice find. Yea this is one open issue with black when you first run on a code base

https://github.com/psf/black/issues/26

tabpy/tabpy_server/handlers/__init__.py
WillAyd: Unfortunately can't do that without exceeding the 88 character line limit, so black makes this a multiline import

PR #385, opened 2019-12-21
Documentation improvements

0golovatyi:
* Update outdated documentation with the links to the latest Tableau pages.

* Use docopt instead of argparser for command line parameters for tabpy and tabpy-user.

* Remove .travis.yml - TabPy uses github actions now.


.github/workflows/test_coveralls_codestyle.yml
```diff
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push, pull_request]
```
johng42: What is the reason for this removal?
According to the docs, both of these are needed for coveralls to run on pull requests.  https://github.com/coverallsapp/github-action/
which isn't intuitive, admittedly...

0golovatyi: As you can see below the build still runs. I removed the `push` to remove duplication of the jobs. In theory, we may have different workflows of push and PR (e.g. only check code style on push) in future. And for now, there's no need to run the workflow twice.

PR #386, opened 2019-12-21
Add return None, modify body args

phantomcosmonaut:
Added the 'return None' capability and unit test. Also modified the code around `arguments_str = ", " + ", ".join(arguments.keys())` not just because I like one-liners, but because arguments.keys() has not actually been sorted, only a new sorted copy has been created, so the arguments_str could potentially combine dict keys out of order.


0golovatyi:
Test failures - https://travis-ci.com/tableau/TabPy/builds/141996187

phantomcosmonaut:
Yep I saw that, I'll fix it

PR #398, opened 2020-01-30
v0.9.0: Fix model deployement with removing usage of pip._internal

0golovatyi:
### Improvements

- Models deployment doesn't depend on pip._internal anymore.
- Package size reduced.


0golovatyi:
> I'm a little confused, if deploying a model just adds a new endpoint and pickled file, can't we just add that to Tabpy itself so that users don't need to deploy anything anymore?

I don't think pickled files are portable, plus deployed models update Tornado config which may have other models in it already and we want to append, not replace Tornado config. 
The better solution would be not to have deploy models with Tornado and `tabpy_tools` but rather install them as packages with `pip`.

0golovatyi:
> Please update the documentation in "Deploying Models Shipped With TabPy"

Thanks for reminding me. Updated.

PR #411, opened 2020-04-14
Secure info

lriggs:
The test changes might be a bit hard to read. I just created a base class that did a lot of the common setup. My first commit has working tests with the duplicated code.


0golovatyi:
The change looks good to me, we need additionally update documentation for the new parameter at https://github.com/tableau/TabPy/blob/secureInfo/docs/server-config.md.

PR #412, opened 2020-04-15
Refactor config parsing to allow custom parsers.

lriggs:
ConfigParser has some helpful functions to parse booleans and ints/floats. The boolean parser handles casing and allows true, 1, yes for true values. This is more relable and user friendly than relying on a direct string match. Updated tests.

One change here is with the TABPY_EVALUATE_TIMEOUT setting. Previously it would try and convert that setting to a float and if it failed would default to 30 seconds. Now it will throw an exception at startup and the process will halt. The new method does handle int -> float conversions.

I could add some extra code to restore the old behavior, but maybe its better this way since the user will know that they have a setting which isn't being used. Mostly silent defaults can be hard to recognize.


tabpy/tabpy_server/handlers/service_info_handler.py
```diff
@@ -15,7 +15,7 @@ def get(self):
         # Some clients may wish to lock down the entire API which can be done through
         # the configuration file.
 
-        if self.settings[SettingsParameters.AuthInfo] == "true":
+        if self.settings[SettingsParameters.AuthInfo] == True:
```
0golovatyi, line 18: Seems like it can be just
`if self.settings[SettingsParameters.AuthInfo]:`

0golovatyi:
@lriggs The coverage for Python code is very flaky, just ignore that -0.06% decline.

PR #425, opened 2020-06-10
Add support for Rserve

0golovatyi:
* With added parameters TabPy can be configured to redirect scripts for Rserve instance.


petr-ponomarenko:
Any updates on this?

0golovatyi:
@petr-ponomarenko It was working for me when I created MR, but it never made to the master branch as with this improvement TabPy is just another layer for the existing topology - it is simply translates HTTP requests to Rserve format. And you need both Rserve and TabPy running. Not sure what is the value of that.

PR #440, opened 2020-08-14
Return HTTP 400 status when receiving a request with authentication credentials and authN is not configured

harold-xi:
When TabPy receives a request with authN credentials, and it it not configured with authentication enabled, it should ignore it and return with a 200 status. Instead, it should return a 400 Bad Request status.

I created another flag authentication_not_required_error that tracks if this specific case happens. I also changed the not_authorized error flag to be authentication_error flag as it now encompasses more than just user not being authorized. The checks in each of the endpoints are also updated to reflect the new scenario and unit tests have been written to test this scenario.

### DIFF
--- a/tabpy/tabpy_server/handlers/base_handler.py
+++ b/tabpy/tabpy_server/handlers/base_handler.py
@@ -6,6 +6,7 @@ import logging
 import tornado.web
 from tabpy.tabpy_server.app.SettingsParameters import SettingsParameters
 from tabpy.tabpy_server.handlers.util import hash_password
+from tabpy.tabpy_server.handlers.util import AuthErrorStates
 import uuid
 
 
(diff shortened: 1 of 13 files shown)


tabpy/tabpy_server/handlers/base_handler.py
```diff
@@ -127,12 +127,14 @@ def initialize(self, app):
         self.password = None
         self.eval_timeout = self.settings[SettingsParameters.EvaluateTimeout]
 
+        self.authentication_not_required_error = None # this is fragile, depends on handle_authentication function because this gets set in the handle_authentication function
```
0golovatyi: Alternatively, you can change `self.not_authorized` to something like `self.auth_error` which can be in one of the states `None`, `NotAuthorized`, and `NotRequired`. With that you won't need to track auth error state with 2 variables.

tabpy/tabpy_server/handlers/endpoint_handler.py
```diff
@@ -22,9 +22,13 @@ def initialize(self, app):
         super(EndpointHandler, self).initialize(app)
 
     def get(self, endpoint_name):
-        if self.should_fail_with_not_authorized():
-            self.fail_with_not_authorized()
-            return
+        if self.should_fail_with_error():
```
0golovatyi, line 25: IMO this check should be as simple as
```
if self.should_fail_with_auth_error():
    self.fail_with_auth_error()
    return
```

0golovatyi:
Could you also change the version to 2.0.0 and update changelist file with the breaking change?

tabpy/tabpy_server/handlers/base_handler.py
```diff
@@ -376,27 +377,36 @@ def handle_authentication(self, api_version) -> bool:
...
 
         if method == "":
-            # Do not validate credentials
-            return True
+            if not self._get_basic_auth_credentials():
+                self.logger.log(logging.DEBUG,
+                    "authentication not required, username and password are none")
+                return AuthErrorStates.NONE
+            else:
+                self.logger.log(logging.DEBUG,
```
0golovatyi, line 397: Maybe make it `logging.ERROR` level

PR #443, opened 2020-08-25
Error loading state: AccessDenied: Access Denied (AWS S3 backend) #18801

harold-xi:
According to Github Issue #18801 Tabpy.query script was not passing headers onto Tabpy, which results in authentication headers not being received and users were unable to use Tabpy.query to query an endpoint. The solution was to just pass on all of the request headers onto Tabpy.


0golovatyi:
@harold-xi seems like some more pep8 complaints need to be fixed.

tabpy/tabpy_server/handlers/util.py
```diff
@@ -2,13 +2,17 @@
...
 from enum import Enum, auto
 
-
 class AuthErrorStates(Enum):
     NONE = auto()
     NotAuthorized = auto()
     NotRequired = auto()
 
 
+class AuthErrorStates(Enum):
```
0golovatyi, line 11: Why this class is duplicated?

harold-xi: Oops, I mixed up my branches.  Will get that fixed right away

jmoens:
all good

PR #444, opened 2020-08-26
Auth bugs fixes

0golovatyi:
### Breaking changes

- TabPy fails with 400 when it is not configure for authentication
  but credentials are provided by client.

### Bug fixes

- When TabPy is running with no console attached it is not failing
  with 500 when trying to respond with 401 status.
- tabpy.query() failing when auth is configured.

### Improvements

- Minor code cleanup.


harold-xi:
Looks good to me

jmoens:
looks good to me

PR #452, opened 2020-09-10
use a new argument in rest_client.py

johng42:
this method accepts a cls value but then immediately reassigns it.  Fix this code smell to make this file a tiny bit easier to maintain by introducing a new variable name


0golovatyi:
Not quite sure what is happening here, need to look at more details how this works...

PR #459, opened 2020-10-07
Deploy to heroku button

ChairmanMa0:
This adds a button to the Readme that deploys TabPy to the user's Heroku account.


0golovatyi (approved):

0golovatyi:
I deployed with the button and the app doesn't work

0golovatyi (changes requested):
After deploying with the button the app doesn't work and can't be reached.

nmannheimer (approved):

ChairmanMa0:
Fixed the issue with the Procfile.
Verified the fix using:
https://heroku.com/deploy?template=https://github.com/tableau/TabPy/tree/deploy-to-heroku-button
This PR can't be tested using the button because it always deploys from master (which of course doesn't have the other changes yet).

PR #471, opened 2021-01-14
Allow to set NLTK options via TabPy config

ElPincheTopo:
Create a convenience function to expose the settings file to scripts, added NLTK options to the example script and updated SentimentAnalysis.py to use the download path and proxy settings defined in the nltk parameters in the config.


0golovatyi:
Looks good. Could you update one of the integration tests for models with the new settings (https://github.com/tableau/TabPy/tree/master/tests/integration)?

ElPincheTopo:
Sure, I'm having a busy week but I hope to do this before the weekend.

cgore1:
@ElPincheTopo would appreciate if you could address integration test comment so that we can merge your change! thanks!

PR #478, opened 2021-02-01
Added authentication for Heroku deployment

illonage:
- Add two config variables for Username + Password
- Add TABPY_PWD_FILE parameter & add an account


0golovatyi:
Looks good.

PR #510, opened 2021-07-08
Add toggle to turn off evaluate API

cgore1:
For security reasons, some clients want to turn off ad hoc python scripts.

Added a new parameter "TABPY_EVALUATE_ENABLE" to TabPy’s configuration file (True by default)
It will serve as a toggle control to TabPy’s evaluate endpoint

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -14,5 +14,6 @@ you may want to consider the following as you use TabPy:
   - Install new Python packages which can contain binary code.
   - Execute operating system commands.
   - Open network connections to other servers and download files.
-- Evaluate endpoint can be disabled by setting "TABPY_EVALUATE_ENABLE" to
-  false in config file
+- Execution of ad-hoc Python scripts can be disabled by turning off the 
+  /evaluate endpoint. To disable /evaluate endpoint, set "TABPY_EVALUATE_ENABLE"
+  to false in config file.
--- a/docs/security.md
+++ b/docs/security.md
@@ -14,6 +14,6 @@ you may want to consider the following as you use TabPy:
   - Install new Python packages which can contain binary code.
   - Execute operating system commands.
   - Open network connections to other servers and download files.
-- Execution of ad-hoc Python scripts can be disabled by turning off the 
+- Execution of ad-hoc Python scripts can be disabled by turning off the
   /evaluate endpoint. To disable /evaluate endpoint, set "TABPY_EVALUATE_ENABLE"
   to false in config file.
--- a/docs/security.md
+++ b/docs/security.md
@@ -14,3 +14,5 @@ you may want to consider the following as you use TabPy:
   - Install new Python packages which can contain binary code.
   - Execute operating system commands.
   - Open network connections to other servers and download files.
+- Evaluate endpoint can be disabled by setting "TABPY_EVALUATE_ENABLE" to
+  false in config file
--- a/docs/server-config.md
+++ b/docs/server-config.md
@@ -87,6 +87,8 @@ at [`logging.config` documentation page](https://docs.python.org/3.6/library/log
 - `TABPY_MAX_REQUEST_SIZE_MB` - maximal request size supported by TabPy server
   in Megabytes. All requests of exceeding size are rejected. Default value is
   100 Mb.
+- `TABPY_EVALUATE_ENABLE` - enable evaluate api. Default
+  value - `true`.
 - `TABPY_EVALUATE_TIMEOUT` - script evaluation timeout in seconds. Default
   value - `30`. This timeout does not apply when evaluating models either
   through the `/query` method, or using the `tabpy.query(...)` syntax with
@@ -124,6 +126,10 @@ settings._
 # specified amount will be rejected by TabPy.
 # Default value is 100 Mb.
 # TABPY_MAX_REQUEST_SIZE_MB = 100
+
+# Toggle for evaluate API
+# Enabled by default. Disabling it will result in 400 error.
+# TABPY_EVALUATE_ENABLE = true
 
 # Configure how long a custom script provided to the /evaluate method
 # will run before throwing a TimeoutError.
--- a/docs/server-config.md
+++ b/docs/server-config.md
@@ -87,8 +87,8 @@ at [`logging.config` documentation page](https://docs.python.org/3.6/library/log
 - `TABPY_MAX_REQUEST_SIZE_MB` - maximal request size supported by TabPy server
   in Megabytes. All requests of exceeding size are rejected. Default value is
   100 Mb.
-- `TABPY_EVALUATE_ENABLE` - enable evaluate api. Default
-  value - `true`.
+- `TABPY_EVALUATE_ENABLE` - enable evaluate api to execute ad-hoc Python scripts
+  Default value - `true`.
 - `TABPY_EVALUATE_TIMEOUT` - script evaluation timeout in seconds. Default
   value - `30`. This timeout does not apply when evaluating models either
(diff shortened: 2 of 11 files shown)


tabpy/tabpy_server/handlers/evaluation_plane_handler.py
```diff
@@ -44,6 +44,10 @@ def initialize(self, executor, app):
 
     @gen.coroutine
     def _post_impl(self):
+        if not self.eval_enabled:
+            self.error_out(400, "Evaluate action is disabled.")
```
lriggs, line 48: Minor point, but I think "endpoint" would be more clear than "action" since that is the term we use in the documentation.

lriggs:
I think it would be good to add a note here: https://github.com/tableau/TabPy/blob/master/docs/security.md

tabpy/tabpy_server/handlers/evaluation_plane_handler.py
rxing-tableau: Or something like, add-hoc script is not allowed. Better confirm with Nathan about the string.

tests/unit/server_tests/test_evaluation_plane_handler.py
```diff
@@ -286,3 +286,89 @@ def test_creds_no_auth_fails(self):
...
+    def tearDownClass(cls):
+        os.remove(cls.state_file.name)
+        os.rmdir(cls.state_dir)
+        os.remove(cls.config_file.name)
+
+    def get_app(self):
+        self.app = TabPyApp(self.config_file.name)
+        return self.app._create_tornado_web_app()
+
+    def test_evaluation_disabled_fails(self):
```
rxing-tableau, line 368: We need test_evaluation_enabled/unset_success as well, cause we do want to support 'add-hoc' script when the flag is true.

cgore1: Done!

tabpy/tabpy_server/handlers/evaluation_plane_handler.py
cgore1: Changed to "Evaluate endpoint is disabled" for now

cgore1: As per Nathan, changed to "Ad-hoc scripts have been disabled on this analytics extension, please contact your administrator."

docs/server-config.md
```diff
@@ -125,6 +127,10 @@ settings._
 # Default value is 100 Mb.
 # TABPY_MAX_REQUEST_SIZE_MB = 100
 
+# Toggle for evaluate API
```
0golovatyi, line 130: You probably want to update https://github.com/tableau/TabPy/blob/master/tabpy/tabpy_server/common/default.conf with the same change.

0golovatyi:
There is no description for the change - what is being changed and why.

0golovatyi:
There is safer and cleaner way for the improvement IMO:
- Add new handler who only returns the error
- On TabPy start up check the flag and use the new handler if the feature is disabled (at https://github.com/tableau/TabPy/blob/master/tabpy/tabpy_server/app/app.py#L153)

With that the check will only be done once and not on every /evaluate request. Plus it won't be possible to make any script to be evaluated until TabPy is restarted with config change.

docs/server-config.md
```diff
@@ -87,6 +87,8 @@ at [`logging.config` documentation page](https://docs.python.org/3.6/library/log
 - `TABPY_MAX_REQUEST_SIZE_MB` - maximal request size supported by TabPy server
   in Megabytes. All requests of exceeding size are rejected. Default value is
   100 Mb.
+- `TABPY_EVALUATE_ENABLE` - enable evaluate api. Default
```
nmannheimer, line 90: I might add a little to the desc here, ie 'enable evaluate api to execute ad-hoc Python scripts'

docs/security.md
```diff
@@ -14,3 +14,5 @@ you may want to consider the following as you use TabPy:
   - Install new Python packages which can contain binary code.
   - Execute operating system commands.
   - Open network connections to other servers and download files.
+- Evaluate endpoint can be disabled by setting "TABPY_EVALUATE_ENABLE" to
```
nmannheimer, line 17: Maybe say something like 'Execution of ad-hoc Python scripts can be disabled by turning off the /evaluate endpoint by...'

nmannheimer (changes requested):

cgore1:
> There is safer and cleaner way for the improvement IMO:
> 
> * Add new handler who only returns the error
> * On TabPy start up check the flag and use the new handler if the feature is disabled (at https://github.com/tableau/TabPy/blob/master/tabpy/tabpy_server/app/app.py#L153)
> 
> With that the check will only be done once and not on every /evaluate request. Plus it won't be possible to make any script to be evaluated until TabPy is restarted with config change.

I am not sure why separate handler would be safer, since in both cases (with or without the separate handler) we will need to restart the app to toggle evaluate api ON/OFF

I feel having inline check for disabled check is simple and effective in this case, we can create a separate handler in future if needs to do extra things. (Again, I am new here, so correct me if I am wrong :) )

0golovatyi:
> I am not sure why separate handler would be safer, since in both cases (with or without the separate handler) we will need to restart the app to toggle evaluate api ON/OFF

It is safer because it is possible to have script or model which deliberately or by mistake modifies the flag in memory (see the first point at https://github.com/tableau/TabPy/blob/master/docs/security.md). Having a handler who just returns the error without even looking at the flag avoids that problem.

cgore1:
> > I am not sure why separate handler would be safer, since in both cases (with or without the separate handler) we will need to restart the app to toggle evaluate api ON/OFF
> 
> It is safer because it is possible to have script or model which deliberately or by mistake modifies the flag in memory (see the first point at https://github.com/tableau/TabPy/blob/master/docs/security.md). Having a handler who just returns the error without even looking at the flag avoids that problem.

Oh interesting! Didn't think of that possibility!
Well, creating a separate handler then

tabpy/tabpy_server/handlers/evaluation_plane_handler.py
```diff
@@ -44,6 +44,11 @@ def initialize(self, executor, app):
 
     @gen.coroutine
     def _post_impl(self):
+        if not self.eval_enabled:
```
rxing-tableau, line 47: Sorry, just got a chance to re-visit.
But can we move this check to 
https://github.com/tableau/TabPy/blob/938ce3b0268435c2738f8b6aa0f30fccd2f82a9d/tabpy/tabpy_server/handlers/evaluation_plane_handler.py#L107
So that we can refuse the request a bit more earlier.
And 404 would be more appropriate I think.

0golovatyi:
One another test you may want to add is for /info handler to return the new setting to a client - https://github.com/tableau/TabPy/blob/master/tests/unit/server_tests/test_service_info_handler.py

cgore1:
> One another test you may want to add is for /info handler to return the new setting to a client - https://github.com/tableau/TabPy/blob/master/tests/unit/server_tests/test_service_info_handler.py

I am trying to add test condition to TestServiceInfoHandlerWithAuth, but the unit test fails with "RuntimeError: Failed to read passwords file ./tests/integration/resources/pwdfile.txt" 

It expects path to pwd file path to be absolute, any way to fix it?
(Changing pwd file path to absolute path in unit test works) I wonder how the current test is successful in first place

nmannheimer (approved):
ship it

lriggs (approved):

PR #516, opened 2021-08-11
Changing error code for AuthErrorStates.NotRequired to 406

jakeichikawasalesforce:
​This change makes the error code distinct for AuthErrorStates.NotRequired, rather than simply a 400 error. This will allow the client (e.g. Tableau) to differentiate between errors more reliably. Adjusted tests accordingly (test_endpoint_handler.py, test_endpoints_handler.py, and test_service_info_handler.py). 

This change also bumps the version to 2.4.0, and makes note of the new toggle to turn off evaluate API.


tabpy/tabpy_server/handlers/base_handler.py
```diff
@@ -434,11 +434,11 @@ def fail_with_auth_error(self):
                 log_message="Invalid credentials provided.",
             )
         else:
-            self.logger.log(logging.ERROR, "Failing with 400 for Bad Request")
-            self.set_status(400)
+            self.logger.log(logging.ERROR, "Failing with 406 for not acceptable")
```
ChairmanMa0, line 437:
```suggestion
            self.logger.log(logging.ERROR, "Failing with 406 for Not Acceptable")
```
nit: I think the names of the HTTP error codes are usually written as all caps

CHANGELOG
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## v2.4.0
+
+### Improvements
+
+- Add toggle to turn off evaluate API.
```
cgore1, line 7: Thanks for adding this 👍

tabpy/tabpy_server/handlers/base_handler.py
jakeichikawasalesforce: Hi there! Thanks for the comment. I was thinking of matching the existing "unauthorized request" format, preceding  on line 428, but I like your suggestion. Let's go with that.

cgore1 (approved):

rxing-tableau (approved):

ChairmanMa0:
@jakeichikawasalesforce Pls tag the release (2.4.0) after it's merged.

PR #518, opened 2021-08-20
Introduce startup script for docker

cgore1:
Fixed bug - tabpy models are not deployed when container is started

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+# Start tabpy
+tabpy &
+
+# Wait for tabpy server to bootup
+sleep 3
+
+# Deploy models
+tabpy-deploy-models &
+
+wait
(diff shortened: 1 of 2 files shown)


start.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# Start tabpy
+tabpy &
+
+# Wait for tabpy server to bootup
```
rxing-tableau, line 6: Anyway we can check if tabpy is actually running? That will be more certain than sleep

cgore1: I thought about this earlier, but it might get unnecessarily complicated.
We can poll on /info api until we get 200 response. To avoid waiting indefinitely, we can add a max retry limit and throw an error I suppose. We also need to know whats the tabpy port number while polling.

PR #534, opened 2022-01-13
Updated python version shield

dleskosky:
Shield shows that only Python 3.7 and 3.8 are now supported


ChairmanMa0 (changes requested):
The right way to fix this is to update the supported Python versions in PyPi (removing v3.6). That will automatically fix the generated image without changing the link.

dleskosky:
> The right way to fix this is to update the supported Python versions in PyPi (removing v3.6). That will automatically fix the generated image without changing the link.

The updated version has been published to PyPi and the shield has been updated automatically.  I am going to close this out.

PR #546, opened 2022-02-28
Add "Support Ukraine" badge with link to National Bank of Ukraine's w…

0golovatyi:
…ebsite


johng42 (approved):

ChairmanMa0:
Hey guys. I'm personally 100% in support of this PR. But Tableau is the official maintainer of the project, I don't think we can jump into this without running it by some folks (lawyers and such).
Need to revert the merge for now. I'll restore the branch as well. I will follow up with SF/Tableau leadership and counsel to see if we can do this.

PR #555, opened 2022-06-08
Remove broken URL

rafaelraposospot:
This URL (http://tabscifi.com/) is not being used so it could be a potential security concern


rafaelraposospot:
Screenshot 2022-06-08 at 16 42 09

0golovatyi:
I used to own that domain, but when I left Tableau, I transferred it to Advanced Analytics team together with backup for the blog. Not sure what happened after that.

jakeichikawasalesforce (approved):

dleskosky (approved):

PR #577, opened 2022-12-19
PROTOTYPE: Add header to static file resonse

lriggs:


dleskosky:
This pull request was created as a template to be used in [this pull request](https://github.com/tableau/TabPy/pull/566) and it can be closed now.

PR #596, opened 2023-03-24
Configurable GRPC port for Arrow Server

antesaj:
- Leverages existing config-file framework to add custom GRPC port capability.
- Will accept via config file, environment variable, and if neither of those are present, fallback to default val.

### DIFF
--- a/tabpy/tabpy_server/app/arrow_server.py
+++ b/tabpy/tabpy_server/app/arrow_server.py
@@ -27,6 +27,9 @@ import uuid
 import pyarrow
 import pyarrow.flight
 
+from tabpy.tabpy_server.app.app_parameters import SettingsParameters
+
+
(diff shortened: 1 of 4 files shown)


tabpy/tabpy_server/app/arrow_server.py
```diff
@@ -148,9 +151,15 @@ def start():
                         help="Enable transport-level security")
     parser.add_argument("--verify_client", type=bool, default=False,
                         help="enable mutual TLS and verify the client if True")
-    parser.add_argument("--config", type=str, default="", help="should be ignored") # TODO: implement config
+    parser.add_argument("--config", type=str, default=None, help="should be ignored") # TODO: implement config
```
antesaj, line 154: Potentially can just derive settings from TabPy?

PR #605, opened 2023-05-03
Fixes incorrect "certificate provided is not valid until ..."

jhilyard:
Fixes https://kb.tableau.com/articles/issue/error-the-certificate-provided-is-not-valid-until-when-starting-tabpy issue: certificate not_before and not_after are naive datetimes in UTC, so the now variable should use datetime.utcnow() not datetime.now().

Currently anyone who installs TabPy and then creates a self-signed certificate to use HTTPS in a time zone west of Greenwich gets an error that the certificate is not yet valid. I found this frustrating enough to fork and provide this fix, since my certificate still won't be valid for over an hour. You can probably also retire the KB article referenced above once merged.

Thanks for your attention!


jakeichikawasalesforce:
@jhilyard thanks for the PR. I think the last step is to sign the "salesforce-cla" then I can merge that for you.

PR #638, opened 2024-06-19
Add TabPy Config Parameter for Minimum TLS Version

jakeichikawasalesforce:
Adding a TabPy config parameter to set the minimum TLS version (TABPY_MINIMUM_TLS_VERSION) that the server will accept for secure connections. By default, the minimum TSL version will be set to TLSv1_2. See https://docs.python.org/3/library/ssl.html#ssl.TLSVersion.MINIMUM_SUPPORTED.

Other changes:
- Fix GitHub Action to use `macos-latest@arm64` in Rest Runs.
- Extend Scrutinzer idle timeout to 10 minutes.
- Use `utcnow()` instead of `now()` when validating certificates (fix for self-signed certificates).


jmoens (approved):

tests/integration/test_minimum_tls_version.py
```diff
@@ -0,0 +1,43 @@
...
+        pwd_file = self._get_pwd_file()
+        if pwd_file is not None:
+            pwd_file = os.path.abspath(pwd_file)
+            config_file.write(f"TABPY_PWD_FILE = {pwd_file}\n")
+
+        config_file.close()
+        self.delete_config_file = True
+        return config_file.name
+
+class TestMinimumTLSVersionValid(TestMinimumTLSVersion):
```
jmoens, line 28: Is there a way to not include the `TABPY_MINIMUM_TLS_VERSION` in the test? In that case should the default get set?

dleskosky (approved):

PR #645, opened 2024-09-06
Add option to specify deployed fuctions as public

jmoens:
This is intended to be used with the tableau Custom Function Explorer. Functions that are specified as public will be shown in tableau with helpful instructions about their usage. Functions with isPublic set to False, or unset, will be available to use, but not viewable in Tableau.

Example function:

With a tabpy server running on 9004, open a terminal, start python and input the following. This will define a function called detectLanguage, as well as a helpful schema that describes its use.

```
from tabpy.tabpy_tools.client import Client

client = Client('http://localhost:9004/')

def detectLanguage(phrase):
    from googletrans import Translator, constants
    translator = Translator()
    detections = []
    for p in phrase:
        print(p)
        detection = translator.detect(p)
        print(detection)
        detections.append(detection.lang)
    print(detections)
    return detections

from tabpy.tabpy_tools.schema import generate_schema

detectLanguageSchema = generate_schema(
    input={'phrase': 'el agua'},
    output='es',
    input_description={'phrase': 'phrase which we would like to detect the language of'},
    output_description='detected language')
```

Next, we deploy the function. Notice that we have a new argument, called isPublic which in this case we have set to true.

`client.deploy('detectLanguage', detectLanguage, 'Detects the language of a given input', schema=detectLanguageSchema, isPublic=True, override=False)`

Here is what we see when we navigate to localhost:9004:
Screenshot 2024-09-06 at 7 41 29 AM

Here is what we see from the tableau Custom Functions Explorer when connected to localhost:9004:
Screenshot 2024-09-06 at 7 42 25 AM

If we would like to update the function to not be visible, we will redeploy: Notice that isPublic is set to False, and override is set to True:

`client.deploy('detectLanguage', detectLanguage, 'Detects the language of a given input', schema=detectLanguageSchema, isPublic=False, override=True)`

Here is what we see when we navigate to localhost:9004:
Screenshot 2024-09-06 at 7 43 13 AM

Here is what we see from the tableau Custom Functions Explorer when connected to localhost:9004:
Screenshot 2024-09-06 at 7 43 40 AM


nmannheimer:
This looks awesome!! The dream is complete :)

jmoens:
@nmannheimer it was a long time coming! Very exciting to see it all come together :)

dleskosky (approved):

PR #652, opened 2024-11-20
Add capability to deploy models remotely

jakeichikawasalesforce:
### Remote Deploy
This PR adds the capability to deploy models to a remote server by creating a client with `remote_server=true`. See this PR's **docs/tabpy-tools.md** for usage.
- Pass *deploy* to *_remote_deploy* if client is connected to a remote server.
- In *_remote_deploy* generate a python script to run on the TabPy server: (i) create client instance, (ii) add credentials if necessary, (iii) add source code, (iv) deploy function with user's parameters.
- Then pass to *_evaluate_remote_script*, which will use the TabPy server's `/evaluate` endpoint to run the script.

If TABPY_EVALUATE_ENABLE is set to False in the config, this remote deployment will not work, which is the intended behavior (otherwise this would provide a hack where the user can upload arbitrary code and then execute it with the query endpoint). 

### Other Fixes:
- Add user's docstring to *_gen_endpoint*, which apparently was not being set at all (always defaulting to "no docstring found"). This uses CustomQueryObject's *get_doc_string*. 
   - However, was running into issues with GitHub's Windows CI/CD Runner, so only fixing this on Mac and Linux for now. This seems to work when ran locally on Windows. Needs more investigation.
- Add `time.sleep(interval)` to beginning of *_wait_for_endpoint_deployment*; this helps prevent a 500 error response when the model did indeed deploy.
- Now deleting an endpoint's stage directory as well, when removing an endpoint.
- Include all local log files in .gitignore.

### DIFF
--- a/tabpy/tabpy_tools/client.py
+++ b/tabpy/tabpy_tools/client.py
@@ -527,7 +527,7 @@ class Client:
 
         msg = response.text.replace('null', 'Success')
         if "Ad-hoc scripts have been disabled" in msg:
-            msg += "\n[Remote TabPy client not allowed.]"
+            msg += "\n[Deployment to remote tabpy client not allowed.]"
 
         status_message = (f"{response.status_code} - {msg}\n")
         print(status_message)


dleskosky (approved):

tabpy/tabpy_tools/client.py
```diff
@@ -447,6 +467,72 @@ def _wait_for_endpoint_deployment(
...
+        response = requests.post(
+            url,
+            headers=headers,
+            auth=self._service.service_client.network_wrapper.auth,
+            json=payload
+        )
+
+        msg = response.text.replace('null', 'Success')
+        if "Ad-hoc scripts have been disabled" in msg:
+            msg += "\n[Remote TabPy client not allowed.]"
```
jmoens, line 530: Maybe Deployment to remote tabpy client not allowed?

jmoens (approved):

PR #666, opened 2026-08-20
Add opt-in OAuth endpoint scopes and bump to 2.15.0

jakeichikawasalesforce:
## Summary

Adds opt-in per-endpoint JWT scopes on TabPy's HTTP path and ships the OAuth work (HTTP #664, Flight #665, and this change) as 2.15.0. This is the last TabPy OSS story in the OAuth epic.

- New config `TABPY_OAUTH_ENFORCE_ENDPOINT_SCOPES` (default `false`).
  When false, existing OAuth installations keep working and tokens without a
  `scope` claim are unchanged.
- When true, after a valid JWT:
  - `/query` requires the configured query scope
  - `/evaluate` requires the configured evaluate scope
  - mutating management operations (`POST /endpoints`,
    `PUT`/`DELETE /endpoints/{name}`, and
    `GET /configurations/endpoint_upload_destination`) require the configured
    deploy scope
- Endpoint scope names are configurable:
  - `TABPY_OAUTH_QUERY_SCOPE` defaults to `tabpy:query`
  - `TABPY_OAUTH_EVALUATE_SCOPE` defaults to `tabpy:evaluate`
  - `TABPY_OAUTH_DEPLOY_SCOPE` defaults to `tabpy:deploy`
- Configurable names support IdP-specific conventions such as Amazon Cognito's
  slash-form scopes (`tabpy/query`, `tabpy/evaluate`, and `tabpy/deploy`).
  Invalid scope tokens fail startup validation. Reusing one scope for multiple
  endpoint groups is allowed but logs a warning.
- Insufficient endpoint scope returns 403 with
  `WWW-Authenticate: Bearer error="insufficient_scope"`. Global
  `TABPY_OAUTH_REQUIRED_SCOPES` misses remain 401.
- `/info`, `/status`, and `GET /endpoints` are not gated by endpoint scopes.
  `/info` advertises the configured query, evaluate, and deploy scope names
  under `oauth-jwt`, even when enforcement is off.
- Nested `tabpy.query()` from `/evaluate` forwards the original token, so it
  requires both the configured evaluate and query scopes.
- Arrow Flight is not per-endpoint scoped and continues to use only
  `TABPY_OAUTH_REQUIRED_SCOPES`. Basic Auth is unaffected.
- Bumps `tabpy/VERSION` to 2.15.0 and documents the OAuth feature in
  `CHANGELOG` and `docs/server-config.md`.

## Testing

- Unit coverage for:
  - path and method scope mapping
  - default and configured scope names
  - configuration validation and `/info` advertisement
  - query, evaluate, and deploy enforcement
  - negative management cases
  - duplicate-scope warnings
  - Basic Auth and global-scope compatibility
- Focused OAuth/JWT/config suites pass on Python 3.13.


tabpy/tabpy_server/handlers/jwt_auth.py
```diff
@@ -354,3 +354,55 @@ def _check_scopes(claims: dict, required_scopes: str) -> None:
...
     if missing:
         raise JwtValidationError(f"JWT missing required scope(s): {', '.join(missing)}")
+
+
+# Well-known endpoint scopes. Bound to HTTP paths when
+# TABPY_OAUTH_ENFORCE_ENDPOINT_SCOPES is true. Not admin-defined.
+SCOPE_QUERY = "tabpy:query"
+SCOPE_EVALUATE = "tabpy:evaluate"
+SCOPE_DEPLOY = "tabpy:deploy"
+WELL_KNOWN_ENDPOINT_SCOPES = (SCOPE_QUERY, SCOPE_EVALUATE, SCOPE_DEPLOY)
```
jmoens: This and `endpoint_scope_for_path` below are both maintaining the scope to endpoint relationship. They are in sync right now, but there is not enforcement for them to remain in sync with later updates. (we don't have any planned currently, but I can see this being an aread of future work!) Did we consider a dict for example? ENDPOINT_SCOPE_MAP = {SCOPE_QUERY = "/query"...} so that there is a single source of truth?

tabpy/tabpy_server/handlers/jwt_auth.py
```diff
@@ -354,3 +354,55 @@ def _check_scopes(claims: dict, required_scopes: str) -> None:
...
+
+_MUTATING_METHODS = frozenset({"POST", "PUT", "DELETE", "PATCH"})
+
+
+def token_has_scope(claims: dict, scope: str) -> bool:
+    """True if `scope` is present in the token's space-separated `scope` claim."""
+    granted = claims.get("scope", "")
+    if not isinstance(granted, str):
```
jmoens, line 390: The implementation here differs slightly from _check_scopes above, in the case of an explicitly null scope. (not unset!) I think that this case is correct, returns false and and then 403 'insufficient scopes', but _check_scopes above will try to split on None

jmoens: Looking more closesly I see that you have addressed this in your PR description. Resolving

tabpy/tabpy_server/app/app.py
```diff
@@ -577,6 +595,44 @@ def _validate_oauth_settings(self):
...
+        ]
+        invalid_scopes = [
+            config_key for setting, config_key in endpoint_scopes
+            if not re.fullmatch(
+                r"[\x21\x23-\x5B\x5D-\x7E]+", self.settings[setting]
+            )
+        ]
+        if invalid_scopes:
+            msg = (
+                f"{', '.join(invalid_scopes)} must be valid, non-empty OAuth "
```
johng42, line 614: Not blocking:  is there a small feature request implied here to parse for what characters are in violation of the regex and present that to the user?  Would be very helpful to point out the (rare) case in which the violating character(s) are hard to figure out.

johng42: or does invalid_scopes get set to the characters in violation?

jmoens (approved):

tests/unit/server_tests/test_oauth_handler.py
```diff
@@ -304,5 +306,346 @@ def test_subject_is_not_logged_by_default(self):
...
+            )
+            configured_deploy_scope = self.fetch(
+                "/configurations/endpoint_upload_destination",
+                headers=self._bearer("tabpy/deploy"),
+            )
+            info = self.fetch("/info", headers=self._bearer("tabpy/query"))
+
+        self.assertEqual(old_scope.code, 403)
+        self.assertNotEqual(configured_scope.code, 401)
+        self.assertNotEqual(configured_scope.code, 403)
```
jmoens: should we assert an actual expected equals here?

johng42 (approved):

tests/unit/server_tests/test_oauth_handler.py
jakeichikawasalesforce: Good call. Updated this to assert 400

tabpy/tabpy_server/app/app.py
jakeichikawasalesforce: Thanks John! Good question. Yeah the invalid_scropes has the affected key names, not the specific characters. Showing the violating characters would be useful though, like for invisible whitespace. I’ll treat that as a non-blocking follow-up.

PR #667, opened 2026-09-15
Redesign the TabPy server landing page

jakeichikawasalesforce:
Redesigns the TabPy landing page with:

- Responsive light and dark themes
- Formatted and JSON views for server information and deployed models
- Model search and expandable metadata
- Improved handling of nested data, loading, and errors
- Updated navigation, resource links, and self-hosted favicon
- Version bump to 2.15.1

This change continues using the existing `/info` and `/endpoints` APIs and does not introduce new endpoints or expose additional information.

**Light and Dark screenshots, no connected to actual server**:
Screenshot_15-9-2026_131516_
Screenshot_15-9-2026_13159_


tabpy/tabpy_server/static/index.html
```diff
@@ -1,71 +1,1107 @@
...
+      const next = currentTheme() === "dark" ? "light" : "dark";
+      document.documentElement.dataset.theme = next;
+      try {
+        window.localStorage.setItem("tabpy-theme", next);
+      } catch (_error) {
+        // The selected theme still applies for this page view when storage is unavailable.
+      }
+      updateThemeControl();
+    });
+    themeMedia.addEventListener("change", function () {
```
jmoens: [Medium] Browser compatibility: This page also uses MediaQueryList.addEventListener, Object.fromEntries, replaceChildren, and multi-argument append without compatibility guards. In unsupported browsers, initialization can fail and leave the page showing permanent loading skeletons. In particular, older Safari supports MediaQueryList.addListener rather than addEventListener. Could we either document the minimum supported browser baseline or add fallbacks for these APIs?

tabpy/tabpy_server/static/index.html
```diff
@@ -1,71 +1,1107 @@
...
+    const body = createElement("div", "model-details");
+    body.append(isObject(model) ? renderObject(model) : renderObject({ value: model }));
+    details.append(summary, body);
+    return details;
+  }
+
+  function populateModelList(list, count, empty, entries, query) {
+    const normalized = query.trim().toLocaleLowerCase();
+    const filtered = entries.filter(function (entry) { return !normalized || flattenForSearch(entry[0], entry[1]).includes(normalized); });
+    list.replaceChildren();
```
jmoens: [Medium] Search performance: Every keystroke serializes the endpoint metadata, destroys the existing model list, and recursively rebuilds the metadata DOM for every matching model, even though the details begin collapsed. This could block the UI on servers with many models or large metadata payloads. Could we precompute each model’s searchable text and either debounce the input or lazily render the detail body when <details> is first opened?

tests/unit/server_tests/test_static_page.py
```diff
@@ -0,0 +1,98 @@
...
+        self.assertIsNotNone(favicon_match)
+        favicon = base64.b64decode(favicon_match.group(1))
+        self.assertEqual(10134, len(favicon))
+        self.assertEqual(
+            "26e7546a015a3299fb76d46030ee9202f6f428ba3505d25734858a279c3b2c54",
+            hashlib.sha256(favicon).hexdigest(),
+        )
+        self.assertNotIn(
+            b"https://www.tableau.com/themes/custom/tableau_www/favicon.ico",
+            page,
```
jmoens: [Low] Test behavior rather than implementation text: These assertions verify that specific source strings exist, but they do not prove that nested rendering, filtering, error handling, or view switching works. They can pass while the UI is broken and fail after harmless refactoring. Could we add a small browser/DOM-level test covering successful rendering, malformed responses, model filtering, formatted/JSON switching, nested authentication metadata, and subdirectory-relative requests?

tabpy/tabpy_server/static/index.html
```diff
@@ -1,71 +1,1107 @@
...
+
+  function errorContent(section, error) {
+    const noun = section === "info" ? "server information" : "deployed models";
+    if (error.status === 401) return createStateCard("!", "Authentication required", "Sign in through your configured TabPy authentication flow to view " + noun + ".", section);
+    if (error.status === 403) return createStateCard("!", "Access not permitted", "Your authenticated session does not have permission to view " + noun + ".", section);
+    if (error.kind === "malformed") return createStateCard("{ }", "Unexpected response", "The server responded, but its " + noun + " payload was not valid JSON in the expected shape.", section);
+    return createStateCard("!", "Unable to load data", "TabPy could not load " + noun + ". Check the server connection and try again.", section);
+  }
+
+  function renderSection(section) {
```
jmoens: [Low, accessibility] Redundant loading content: Loading is already announced by the adjacent role="status" live region. The visual skeleton and its empty child elements do not convey additional information and should be hidden from assistive technology. Could we set aria-hidden="true" on the skeleton container and rely on the existing live-region message?

tests/unit/server_tests/test_static_page.py
jakeichikawasalesforce: Removed the brittle implementation-string assertions. Since the project does not currently have browser/DOM test infrastructure, I’ve kept interactive verification manual rather than introducing a new test framework in this PR.

jmoens (approved):

